PR #3, opened 2026-02-19
Server side architecture

### LINKED ISSUE #1: Server-side architecture
### Summary

This issue proposes a discussion on the server-side architecture to better handle high-performance scenarios and avoid potential bottlenecks. The current architecture, which funnels all incoming packets through a single `quic_listener` process, could limit scalability.

Having a look at `src/quic_listener.erl`, the current server design operates as follows:
1. A single `gen_server` process (`quic_listener`) binds to a UDP port.
2. This one process receives *all* incoming UDP datagrams for all connections.
3. For each datagram, it decodes the header to find the Destination Connection ID (DCID).
4. It then looks up the corresponding connection process PID in an ETS table and forwards the packet.

This makes the `quic_listener` a centralized dispatcher.

### The Performance Bottleneck

While this design is clean and effective for a certain level of traffic, it has an inherent scalability limit. The single dispatcher process is a serialization point and under heavy packet load (e.g., from many concurrent connections or high-bandwidth streams), this process can become a bottleneck, preventing the system from fully utilizing available CPU cores.

### Proposed Solution: A Pool of Listeners

To address this, we could explore moving to a **listener pool** architecture. The core idea is:

1. **Multiple Listener Processes:** Instead of one, start a pool of `quic_listener` processes (e.g., one per scheduler or core or some proportion of it).
2. **Shared Port:** Each process would open its own socket but bind to the *same* port and IP address, using the `reuseport` socket option.
3. **Kernel-Level Load Balancing:** With `reuseport`, the OS kernel distributes incoming UDP datagrams across the pool of listening sockets. This provides efficient, low-level load balancing.

This would parallelize the work of receiving and dispatching packets, allowing the server to scale much more effectively with the number of available CPU cores.

There's also the point where the listener does `active, true`. This is good for performance, until you can get attacked at that port. We should be doing some `active, N` with N somewhere between 50 to a 100, in my benchmarks that's usually the sweet-spot for performance.

There's a myriad other problems regarding being attacked on ports, that I've extensively worked on in DNS (see `erldns` repo, sorry for self-promotion). We might discuss implementing similar things here if desired, but that's probably material for a separate issue.

Looking forward to hearing some thoughts :)

---

And above all, @benoitc, THANK YOU. I've been wanting to have a pure-BEAM option for QUIC since I ever heard QUIC existed. MASSIVE KUDOS!

NelsonVides:
See the commit messages for a lot of details, hopefully that's helps with the review very well.

This is probably enough to close #1 as I had it in mind, but I might propose in follow-up tickets more things regarding DDoS protections, UDP is not that easy protocol as TCP is 🥲 

I also need to discover a bit better how tests work and ensure better testing for my changes, these are kinda just a refactoring of equivalent behaviour but in any case, more or less 🤔 

PS: @benoitc I've written you an email too. Looking forward to help out with the server side of things here :)

### DIFF
--- a/src/quic_listener.erl
+++ b/src/quic_listener.erl
@@ -50,6 +50,7 @@
 %% gen_server callbacks
 -export([
     init/1,
+    handle_continue/2,
     handle_call/3,
     handle_cast/2,
     handle_info/2,
@@ -68,6 +69,7 @@
     alpn_list :: [binary()],
     %% Connection ID -> Pid mapping
     connections :: ets:tid(),
+    tickets_table :: ets:tid(),
     %% Stateless reset secret (RFC 9000 Section 10.3)
     reset_secret :: binary(),
     %% Connection handler callback: fun(ConnPid, ConnRef) -> {ok, HandlerPid}
@@ -75,6 +77,7 @@
     %% Options
     opts :: map()
 }).
+-type state() :: #listener_state{}.
 
 %%====================================================================
 %% API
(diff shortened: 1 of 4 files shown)


benoitc:
Sorry I didn't saw your PR (will check the mail) until now. Can you rebase your changes against latest master?

benoitc:
Hey Nelson, thanks for the PR!

I looked at it carefully. The changes are interesting but I have some concerns:

- The `reuseport_lb` thing only works on Linux (it's a kernel feature). On macOS/BSD it won't help.

- Cowboy calls `quic_listener:start/2` directly, not the pool supervisor. So it won't benefit from the pool improvements anyway.

- Adding 2 new modules + deeper supervision for something that only helps in a specific case (Linux + pool mode) feels a bit heavy.

Maybe we could do this differently? Like writing a helper or an example showing how to setup a high-perf listener pool on Linux? Keep the core simple for the common case.

The `ets:select_delete` change is nice though - happy to take that one separately if you want to send a smaller PR.

What do you think?

NelsonVides:
Updated.

Btw, I've also been careful to avoid named ets tables or processes, so that there's no global naming collision and we're not prevented from spawning more than one listener in a server. Also makes it possible to run tests in parallel, if they share no state :)

benoitc:
Thank you! This has been merged in main at 11eac745e3bfde787c14a3ea5ee9582841e5f022.

PR #11, opened 2026-02-25
Structured logging

NelsonVides:
1. Stop using `error_logger`, it was deprecated like, almost a decade ago.
2. Use structured logging and not flat strings, this is library code, not application, we never know that logging structure the parent application of this library will need and there's nothing less structured than opaque strings. We pass a data-structure, which in this case is a map, and let the parent application decide how to format what, and even how to filter things and ignore logs its not interested in, because with a structure it can do such things.
3. Offer a `report_cb` callback for the case when the user is interested in a human-readable string. This callback is optional, and it takes the logs we generate and knows how to transform them into a human-friendly string. The callback is part of the OTP Logger API and is intended for custom formatters to optionally call it. `supervisor` [implements such thing](https://github.com/erlang/otp/blob/b74eae92e1dee97d57a9ab3ea3c36e8b701a9ca6/lib/stdlib/src/supervisor.erl#L2296) for example.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -32,6 +32,10 @@
 -behaviour(gen_statem).
 
 -include("quic.hrl").
+-include_lib("kernel/include/logger.hrl").
+-define(QUIC_LOG_META, #{
+    domain => [erlang_quic, connection], report_cb => fun quic_log:format_report/2
+}).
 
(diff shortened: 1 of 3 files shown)


NelsonVides:
Added domains too. See https://www.erlang.org/doc/apps/kernel/logger.html#t:metadata/0 for details.

PR #53, opened 2026-04-14
fix(test): start quic in h3spec, skip docker-gated suites, make targets

benoitc:
Three small fixes after running the full local test matrix on main.

`quic_h3_h3spec_SUITE:init_per_suite` only started `crypto` and `ssl`, so `quic:start_server/3` inside the test body crashed on `ets:lookup(quic_server_registry_tab, ...)`. Matches the pattern already used in the other suites.

The e2e/bbr/cubic/h3-e2e suites all probe their Docker echo server via `gen_udp:send`, which always succeeds locally because UDP is connectionless. Running `rebar3 ct` without `docker compose up` booted the suites against whatever happened to be on port 4433 and then timed out with opaque errors. Gated `init_per_suite` on `QUIC_SERVER_HOST` / `H3_SERVER_HOST` being explicitly set; otherwise the suite returns a `{skip, ...}` with a reason that points at docker compose. CI that exports the variable is unaffected.

Added `make test-local`, `make test-docker`, and `make test-all` so contributors have one command each for the two stages:

```
make test-local    # eunit, proper, in-process CT
make test-docker   # docker compose up, external-server CT, compose down
make test-all      # both plus xref/lint/dialyzer
```

`make test-local` currently green: 1916 eunit + 74 PropEr properties + 83 CT cases across five suites.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,3 +22,44 @@ erlang.mk:
 endif
 
 include $(if $(ERLANG_MK_FILENAME),$(ERLANG_MK_FILENAME),erlang.mk)
+
+##
+## Test convenience targets layered on top of rebar3.
+##
+## `make test-local` runs everything that doesn't need Docker.
+## `make test-docker` spins the docker-compose echo + H3 servers up,
+## points the external-server CT suites at them, and tears them down.
+## `make test-all` runs both stages followed by static checks.
+##
+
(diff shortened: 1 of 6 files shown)


benoitc:
Superseded by #54 (in-process e2e servers) and #55 (h3spec app start + make test targets).

PR #142, opened 2026-05-25
fix(qpack): correct field-section prefix Base and dynamic-table encoding (RFC 9204)

williamthome:
The QPACK Encoded Field Section Prefix and the dynamic table did not follow RFC 9204. Because of that, nghttp3 rejected the encoder output and the dynamic table did not work. The library's own decoder is not strict, so the bug stayed hidden. The dynamic path also had no round-trip tests. This PR fixes the prefix Base (encode and decode), the Insert With Literal Name opcode, the dynamic-table Base frame, and the Required Insert Count integer width. It also adds round-trip tests and nghttp3 interop tests.

Example, static `:status: 200`: `00 80 D9` becomes `00 00 D9`. The old Base byte `0x80` means Base = -1, which a strict decoder rejects.

The main rule is RFC 9204 §4.5.1.2 (Base):

> A Sign bit of 0 indicates that the Base is greater than or equal to the value of the Required Insert Count; the decoder adds the value of Delta Base to the Required Insert Count to determine the value of the Base. A Sign bit of 1 indicates that the Base is less than the Required Insert Count; the decoder subtracts the value of Delta Base from the Required Insert Count and also subtracts one [...]. That is:
>
> ```
> if Sign == 0: Base = ReqInsertCount + DeltaBase
> else:         Base = ReqInsertCount - DeltaBase - 1
> ```
>
> An encoder that produces table updates before encoding a field section might set Base to the value of Required Insert Count. In such a case, both the Sign bit and the Delta Base will be set to zero.

So a static-only section (Required Insert Count 0, Base 0) is `00 00`. A dynamic section signals `Base = insert_count` with `DeltaBase = insert_count - RIC`. Related parts: §4.5.1.1 (the Required Insert Count is an 8-bit prefix integer, not a raw byte) and §4.3.3 (Insert With Literal Name is `01Hxxxxx`).

Verified against nghttp3: `h2load --alpn-list=h3` went from 0/200 to 200/200, and the qifs dynamic fixtures (capacity 256 and 4096) now decode correctly.

These bugs came up while adding HTTP/3 to roadrunner (arizona-framework/roadrunner#20), which builds on erlang_quic.

Thanks for the library, @benoitc! I am happy to change anything here if you prefer a different approach.

### DIFF
--- a/src/qpack/quic_qpack.erl
+++ b/src/qpack/quic_qpack.erl
@@ -156,28 +156,34 @@ new(Opts) ->
 %% @doc Encode headers using QPACK with state.
 -spec encode(headers(), state()) -> {binary(), state()}.
 encode(Headers, State) ->
-    %% First pass: encode headers and track max dynamic table index referenced
-    {EncodedHeaders, NewState, MaxRefIndex} = encode_headers_tracking(Headers, State, <<>>, -1),
+    %% Base is the reference frame for the dynamic relative indices in this
+    %% field section (RFC 9204 Section 3.2.6). The field lines below encode
+    %% each RelIndex relative to the insert count at section start, so snapshot
+    %% it and signal it as the Base. Entries referenced here were inserted (and
+    %% acknowledged) before this section, so their absolute index < BaseIC; the
+    %% running insert count must not be used, as inserts during this section
+    %% would shift the frame between field lines.
+    BaseIC = State#qpack.insert_count,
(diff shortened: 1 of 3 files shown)


benoitc:
Merged, thanks for the fix and the clear RFC 9204 references @williamthome

benoitc:
fyi 1.4.3 has been cut .

PR #148, opened 2026-05-29
Add 0-RTT (early data) support

### LINKED ISSUE #147: Missing 0-RTT (early data) support
Hello, i'm working on [Quiver](https://github.com/edlontech/quiver) to implement HTTP/3 Client support and got stuck with early-data support, right now there's no 0-RTT on quicer, so resumed connections still pay a full round trip before they can send data. The lib should let a client with a session ticket send application data in its first flight and have the server accept it.

Roughly what needs doing:

- Client: derive early keys from a stored ticket and send data using the early_data extension (TLS 1.3 — RFC 8446 §2.3, §4.2.10), carried over QUIC 0-RTT packets per RFC 9001 §4.6.
- Server: validate the resumption PSK + binder (RFC 8446 §4.2.11), accept or reject early data, and honor max_early_data_size from the ticket (RFC 8446 §4.6.1). On reject, the client just resends at 1-RTT.
- Replay protection: 0-RTT isn't forward-secret and can be replayed, so consume tickets on first use (RFC 9001 §9.2).
- HTTP/3: a client sending requests in early data has to reuse the server's remembered SETTINGS until the new ones arrive (RFC 9114 §7.2.4.2).

ycastorium:
Lets a client with a session ticket send application data in its first flight instead of paying a full round trip on resumption. Wires the existing TLS 1.3 / QUIC PSK machinery through the public API and the HTTP/3 layer.

**What's in here**

Public API (quic.erl, quic_connection.erl)
- New accessors quic:has_early_keys/1 and quic:early_data_accepted/1. The first reports whether early keys were derived from a stored ticket (0-RTT is possible); the second reports whether the server accepted early data, or unknown before the handshake settles.

TLS (quic_tls.erl)
- Parse the server's acceptance signal — an empty early_data extension echoed in EncryptedExtensions; presence is the whole signal.

HTTP/3 (quic_h3.erl, quic_h3_connection.erl)
- New bootstrapping and early_data states. After ownership transfer the client H3 process is primed and routes to either the 0-RTT early_data path or the regular awaiting_quic wait based on quic:has_early_keys/1.
- quic_h3:early_data_accepted/1 exposes the outcome at the H3 layer.
- Early-data requests reuse the server's remembered SETTINGS until the new SETTINGS arrive

Replay
- 0-RTT data isn't forward-secret and is replayable, so resumption tickets are consumed on first use.

I added meck as a dep, but i can remove it if needed.

Closes #147

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -17,7 +17,7 @@
     {test, [
         {deps, [
             {proper, "1.4.0"},
-            {meck, "1.0.0"}
+            {meck, "1.2.0"}
         ]},
         {erl_opts, [debug_info, nowarn_export_all]}
     ]},
--- a/rebar.config
+++ b/rebar.config
@@ -16,7 +16,8 @@
 {profiles, [
     {test, [
         {deps, [
-            {proper, "1.4.0"}
+            {proper, "1.4.0"},
+            {meck, "1.0.0"}
         ]},
         {erl_opts, [debug_info, nowarn_export_all]}
     ]},
--- a/src/h3/quic_h3_connection.erl
+++ b/src/h3/quic_h3_connection.erl
@@ -485,7 +485,12 @@ init({client, QuicConn, _Host, _Port, Opts, Owner}) ->
         owner = Owner,
         owner_monitor = MonRef,
         local_settings = LocalSettings,
-        qpack_encoder = quic_qpack:new(#{max_dynamic_size => MaxTableCapacity}),
+        %% Encoder starts static: its dynamic table is enabled only once the
+        %% peer's SETTINGS are applied (apply_peer_settings/2), never from our
+        %% local settings, so 0-RTT early-data requests stay static-only until
+        %% the server's remembered capacity is known (RFC 9114 §7.2.4.2). The
+        %% decoder is sized by the capacity we advertise, which we must honour.
+        qpack_encoder = quic_qpack:new(#{max_allowed_capacity => MaxTableCapacity}),
         qpack_decoder = quic_qpack:new(#{max_dynamic_size => MaxTableCapacity}),
         % Client uses even stream IDs (0, 4, 8, ...)
         next_stream_id = 0,
@@ -524,7 +529,12 @@ init({server, QuicConn, Opts, Owner}) ->
         owner = Owner,
         owner_monitor = MonRef,
         local_settings = LocalSettings,
-        qpack_encoder = quic_qpack:new(#{max_dynamic_size => MaxTableCapacity}),
+        %% Encoder starts static: its dynamic table is enabled only once the
+        %% peer's SETTINGS are applied (apply_peer_settings/2), never from our
+        %% local settings, so 0-RTT early-data requests stay static-only until
+        %% the server's remembered capacity is known (RFC 9114 §7.2.4.2). The
+        %% decoder is sized by the capacity we advertise, which we must honour.
+        qpack_encoder = quic_qpack:new(#{max_allowed_capacity => MaxTableCapacity}),
         qpack_decoder = quic_qpack:new(#{max_dynamic_size => MaxTableCapacity}),
         % Server uses odd stream IDs (1, 5, 9, ...)
         next_stream_id = 1,
--- a/src/h3/quic_h3_connection.erl
+++ b/src/h3/quic_h3_connection.erl
@@ -589,6 +589,13 @@ bootstrapping(
 ) ->
     {keep_state, forward_early_data_rejected(StreamIds, State)};
 bootstrapping(
+    info,
+    {quic, QC, {session_ticket, T}},
+    #state{quic_conn = QC} = State
+) ->
+    ok = forward_session_ticket(T, State),
+    keep_state_and_data;
+bootstrapping(
     {call, From},
     early_data_accepted,
     #state{quic_conn = QC} = _State
--- a/src/h3/quic_h3_connection.erl
+++ b/src/h3/quic_h3_connection.erl
@@ -20,6 +20,7 @@
 -export([
     start_link/3,
     start_link/4,
+    prime/1,
     request/2,
     request/3,
     open_bidi_stream/2,
@@ -61,7 +62,9 @@
 
 %% State functions
 -export([
+    bootstrapping/3,
     awaiting_quic/3,
+    early_data/3,
     h3_connecting/3,
     connected/3,
     goaway_sent/3,
@@ -258,7 +261,14 @@
     %% Final-response HEADERS frames buffered until the first body chunk so
     %% they coalesce into one QUIC packet (StreamId -> encoded HEADERS frame).
     %% Placed at the end so prior tuple positions stay stable for tests.
-    pending_response_headers = #{} :: #{stream_id() => binary()}
+    pending_response_headers = #{} :: #{stream_id() => binary()},
+
+    %% Set in bootstrapping(cast, prime, ...) from quic:has_early_keys/1.
+    has_early_keys = false :: boolean(),
+
+    %% Set when {quic, _, {connected, _}} arrives. Used by early_data
+    %% convergence logic to detect when 1-RTT is up.
+    quic_connected = false :: boolean()
 }).
 
 %%====================================================================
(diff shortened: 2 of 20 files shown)


test/quic_h3_0rtt_SUITE.erl
```diff
@@ -0,0 +1,622 @@
...
+            drain_exits(),
+            process_flag(trap_exit, OldTrap)
+        end,
+    case Outcome of
+        {ok, EarlyData, Status} ->
+            ct:pal(
+                "Resumed in-proc: early_data_accepted=~p status=~p",
+                [EarlyData, Status]
+            ),
+            ?assert(lists:member(EarlyData, [true, false, unknown])),
```
benoitc: early_data_accepted/1's entire range is true|false|unknown, so it can never fail . This doesn't prove the server accepted 0-RTT.  Rather add a path asserting That EarlyData is tru and was processed before handshake completion

src/quic_connection.erl
```diff
@@ -5102,16 +5155,29 @@ process_tls_message(
...
     case quic_tls:parse_encrypted_extensions(Body) of
-        {ok, #{alpn := Alpn, transport_params := TP}} ->
+        {ok, #{alpn := Alpn, transport_params := TP, early_data := EarlyDataInEE}} ->
+            %% RFC 9001 §4.6.2: a client that offered 0-RTT (i.e. has
+            %% derived early_keys from a session ticket) confirms
+            %% acceptance by the presence of the empty `early_data'
+            %% extension in EncryptedExtensions. Absence is rejection.
+            %% Clients that did not offer 0-RTT keep the default `false'.
+            OfferedZeroRtt = State#state.early_keys =/= undefined,
+            EarlyDataAccepted = OfferedZeroRtt andalso EarlyDataInEE,
```
benoitc, line 5259: good

src/quic_connection.erl
```diff
@@ -7350,6 +7420,48 @@ send_zero_rtt_packet(Payload, EarlyKeys, State) ->
...
+        #{
+            what => zero_rtt_rejected,
+            rejected_stream_ids => RejectedIds,
+            count => length(RejectedIds)
+        },
+        ?QUIC_LOG_META
+    ),
+    Owner ! {quic, self(), {early_data_rejected, RejectedIds}},
+    State#state{
+        streams = NewStreams,
```
benoitc: If the application re-issues the same bytes on new 1-RTT streams, the connection-level bytes-sent counter still includes the discarded  0-RTT bytes this can trigger premature MAX_DATA exhaustion. Shouldn't we  reset send-side FC to the server's real transport params.? Also: do  rejected stream IDs leave a gap the server (which never saw them) handles cleanly?

src/quic_connection.erl
```diff
@@ -175,7 +178,14 @@
...
     test_maybe_send_ack_app/2,
-    test_classify_recv_trigger/2
+    test_classify_recv_trigger/2,
+    %% 0-RTT rejection (RFC 9001 §4.6.2)
+    test_state_for_zero_rtt_reset/3,
+    test_zero_rtt_reset_info/1,
+    test_finalize_zero_rtt_handshake/2,
+    test_reset_zero_rtt_streams/2,
+    test_client_state_for_ee/4,
+    test_process_encrypted_extensions/2
```
benoitc: Project convention is to keep test-only exports in a separate module, can you move them out in test folder ?

src/h3/quic_h3_connection.erl
```diff
@@ -530,6 +552,172 @@ terminate(_Reason, _StateName, #state{quic_conn = QuicConn}) ->
...
+
+%%====================================================================
+%% State: early_data
+%% Open critical H3 streams and send SETTINGS as 0-RTT data. Accept
+%% request calls and forward HEADERS as 0-RTT. Converge to connected
+%% when both QUIC has handshaked AND peer SETTINGS has arrived.
+%% RFC 9114 §6.2.1 (SETTINGS-first), §7.2.4.2 (0-RTT initialization).
+%%====================================================================
+
+early_data(enter, _OldState, State) ->
```
benoitc, line 636: RFC 9114 §7.2.4.2: early requests must respect the server's remembered SETTINGS . is this loaded/ applied to qpack encoding? will it default to static or defuaults?

src/h3/quic_h3_connection.erl
```diff
@@ -530,6 +552,172 @@ terminate(_Reason, _StateName, #state{quic_conn = QuicConn}) ->
...
+%% chance to cast `prime'. Without postponing, the catch-all clause
+%% below would silently drop these and strand the H3 FSM in a state
+%% that waits for them to arrive again. RFC 9114 §3.3 ordering.
+bootstrapping(info, {quic, QC, {connected, _}}, #state{quic_conn = QC}) ->
+    {keep_state_and_data, [postpone]};
+bootstrapping(info, {quic, QC, {stream_data, _, _, _}}, #state{quic_conn = QC}) ->
+    {keep_state_and_data, [postpone]};
+bootstrapping(info, {quic, QC, {new_stream, _, _}}, #state{quic_conn = QC}) ->
+    {keep_state_and_data, [postpone]};
+bootstrapping(_EventType, _Event, _State) ->
```
benoitc, line 625: bootstraping handle early_data_rejected but not session_tickett likeoter states does. Maybe we can make it consistent there too ?

rebar.config
```diff
@@ -16,7 +16,8 @@
 {profiles, [
     {test, [
         {deps, [
-            {proper, "1.4.0"}
+            {proper, "1.4.0"},
+            {meck, "1.0.0"}
```
benoitc: bump to latest to support erlang 29.

benoitc (changes requested):
CI is red for now can you take care of it ? also the branch predates recent changes to support IPv6 can upi ensure to do your changes over the last main ? 

i've commented the other parts in the source.

ycastorium:
> CI is red for now can you take care of it ? also the branch predates recent changes to support IPv6 can upi ensure to do your changes over the last main ?
> 
> i've commented the other parts in the source.

Sure! I'll check!

src/quic_connection.erl
ycastorium: The 0-RTT send bytes are tracked in a separate early_data_sent counter, the 0-RTT send path never adds them to the connection-level data_sent. On acceptance they're folded into data_sent and on rejection (reset_zero_rtt_streams) early_data_sent is zeroed and data_sent is left untouched, and then the 1-RTT keys bring a fresh initial_max_data from the server's real transport params, so the re-issued requests count against data_sent from zero.
 
On the stream IDs: rejected streams are dropped locally and no RESET_STREAM is sent (the server discarded the 0-RTT packets and never opened them). The app re-issues on fresh 1-RTT stream IDs; the skipped low IDs are implicitly opened server-side per RFC 9000 3.2 and sit idle i think this is benign? 

I did found some other bugs though, fixed them.

src/h3/quic_h3_connection.erl
ycastorium: Woops, hood catch. By default it's safe, i checked and H3 default qpack_max_table_capacity is 0, so the encoder is static-only and 0-RTT HEADERS are static+literals. But the encoder was sizing its dynamic table from our local settings at init, so a non-zero local capacity would have made early data use the dynamic table before the server's real/remembered capacity was known. 

It works safely now, but persisting server SETTINGS to enable dynamic-table 0-RTT will need more work and i don't think its useful for now.

src/h3/quic_h3_connection.erl
ycastorium: Done, bootstrapping now forwards session_ticket the same way the other states do.

rebar.config
ycastorium: Updated!

test/quic_h3_0rtt_SUITE.erl
ycastorium: It was worse than that, by running against aioquic i found quite some issues, i dropped the useless test and added a suite to test against aioquic instead.

benoitc:
merged . thanks for the patch! 1.6.1 is now available

PR #161, opened 2026-06-03
Fix H3 connect race: forward QUIC events leaked to caller before transfer

### LINKED ISSUE #160: HTTP/3 client quic_h3:connect/3 hangs until connect_timeout on fast handshakes
Currently, quic_h3:connect/3 opens the QUIC connection with the calling process as owner, then transfers ownership to the H3 connection process via quic:set_owner_sync/2. On fast handshakes (localhost, low-RTT peers, 0-RTT resumption), the QUIC connected event and the server's control-stream data carrying SETTINGS are delivered to the caller's mailbox during the brief window before that transfer completes. set_owner_sync/2 re-emits connected to the new owner, so that event is recovered, but it cannot replay the already-delivered stream_data messages. Those SETTINGS events are stranded in the caller's mailbox, the H3 state machine never sees them, settings_received never becomes true, and the connection never converges.

The bug is intermittent because it depends purely on whether the handshake completes faster than the few microseconds of start_link + set_owner_sync: slow peers win the race and work, fast/low-latency/resumed handshakes lose it deterministically.

ycastorium:
Fixes #160

### DIFF
--- a/src/h3/quic_h3.erl
+++ b/src/h3/quic_h3.erl
@@ -278,6 +278,15 @@ start_h3_connection(QuicConn, HostBin, Port, Opts) ->
         {ok, H3Conn} ->
             %% Transfer ownership to H3 process so it receives QUIC events
             ok = quic:set_owner_sync(QuicConn, H3Conn),
+            %% The QUIC connection was opened with this (caller) process as
+            %% owner, so any owner events emitted before set_owner_sync took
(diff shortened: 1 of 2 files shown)


benoitc:
Superseded by #162, which incorporates this same `quic_h3:connect/3` fix plus the `quic_happy` race-coordinator handoff, and ports your regression test (credited). Thanks @ycastorium.

PR #173, opened 2026-06-23
fix: connection flow control sliding window bug

maslowalex:
While working on media over QUIC feature, I noticed that connection stops transmitting when traffic hits 8MB, digging deeper I discovered that the window calculation has a bug! Please take a look at this.


benoitc:
Same fix landed in #176 (merged). Thanks for catching the connection-flow-control stall at 8 MiB.

PR #174, opened 2026-06-24
fix: drive idle timer from received activity, not sends (RFC 9000 §10.1)

sstrollo:
quic_connection reset `last_activity` on every successful 1-RTT packet send. Because the idle timer and keep-alive are both derived from `last_activity`, an endpoint that kept sending into a black hole — e.g. a client emitting keep-alive PINGs and PTO retransmits while its peer is unreachable — perpetually reset its own idle timer and never timed out a dead path. The connection stayed "up" forever, so application-level failover (e.g. QUIC -> TCP) that waits for the connection to close never triggered.

RFC 9000 §10.1 restarts the idle timer when a packet is *received*, and only on send for the first ack-eliciting packet since the last receive. The simplest correct behaviour is to drive the idle timer purely from received activity: a healthy peer ACKs our packets, and those received ACKs are what keep the timer alive; a dead path produces no ACKs and so idle-closes as intended. Keep-alive PINGs continue to keep the *peer's* timer from firing.

Fix: stop updating `last_activity` on the send path (send_app_packet_internal). It is already updated for every received packet (decrypt_packet) and at handshake completion, which is the correct semantics for both the idle timer and keep-alive.

Add a behavioural regression test (quic_idle_timeout_tests): a client connected over a controllable in-process datagram bridge keeps sending keep-alive PINGs while the path is black-holed and must still idle-close (fails before this fix), plus a live-path test asserting keep-alive holds a healthy connection open (guards against over-aggressive closing).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -3541,13 +3541,19 @@ send_app_packet_internal(Payload, Frames, State) ->
                     undefined -> State#state.socket_state;
                     _ -> NewSocketState
                 end,
+            %% Note: `last_activity' is deliberately NOT updated here. It tracks
+            %% activity *received from the peer* and drives the idle timer
+            %% (RFC 9000 §10.1) and keep-alive. Bumping it on every send meant a
+            %% peer that kept sending keep-alive PINGs / PTO retransmits into a
+            %% black hole reset its own idle timer forever and never timed out a
+            %% dead path. On a healthy path the peer's ACK is a received packet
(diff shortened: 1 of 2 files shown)


sstrollo:
Hi Benoit. Great that you started this project! While testing a bit with our code base (that currently is using quicer) I ran into the issue that we had a test that would block traffic and test that the connection timed out, which it didn't. Claude found this root cause and wrote a test for it.

PR #178, opened 2026-06-26
server: validate the client certificate chain (mutual TLS)

sstrollo:
With verify => true the server requested a client certificate and checked the CertificateVerify signature (proof the client holds the leaf's private key), but never validated the certificate chain against the configured trust anchors and silently accepted an empty certificate. A client could therefore authenticate with a self-signed certificate carrying any subject it liked, or with none at all — the TLS layer provided no CA-backed client authentication.

Add quic_cert:validate_client/3 (the same trust-anchor chain validation as validate_server/4, without the hostname check, since a client certificate is not bound to a server name) and call it when the server receives the client's Certificate. Reject an unvalidated chain (bad_certificate / unknown_ca) and an empty certificate (certificate_required) instead of proceeding to Finished.

Tests: quic_cert:validate_client accepts a CA-issued chain and rejects a self-signed "faked" cert (unknown_ca), a missing cert, and an empty trust store. A before/after test shows the CertificateVerify signature on a faked self-signed cert still verifies (what the server accepted on before) while chain validation now rejects the same cert.

### DIFF
--- a/src/quic.erl
+++ b/src/quic.erl
@@ -767,6 +767,14 @@ get_peer_transport_params(Conn) when is_pid(Conn) ->
 %% <ul>
 %%   <li>`cert' - DER-encoded certificate</li>
 %%   <li>`key' - Private key term</li>
+%%   <li>`verify' - Request a client certificate (mutual TLS) and validate any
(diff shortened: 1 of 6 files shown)


sstrollo:
Code seems okay, please /rerun the tests

Issue #134, opened 2026-05-20
I found an interoperability bug with Chromium (v150 dev/canary) during QUIC + TLS 1.3 handshake.


kreactnative:
send_server_handshake_flight sends the entire TLS handshake flight (EncryptedExtensions + Certificate chain + CertificateVerify + Finished, typically 3–5 KB) as a single UDP datagram, violating the QUIC RFC requirement to respect max_udp_payload_size.

Any QUIC client that strictly enforces its advertised max_udp_payload_size (e.g. Chromium, which advertises 1472 bytes) drops the oversized packet with ERR_MSG_TOO_BIG. The TLS handshake stalls and the client closes the connection with QUIC_NETWORK_IDLE_TIMEOUT after ~4 seconds. curl works by accident because its UDP socket buffer is generous.

PR #195, opened 2026-08-09
Add X25519MLKEM768 post-quantum hybrid key exchange

jbevemyr:
This adds the hybrid ML-KEM-768 + X25519 named group (draft-ietf-tls-ecdhe-mlkem, code point `0x11EC`) as an opt-in key-exchange group, making erlang_quic able to negotiate quantum-safe key exchange with any modern peer (OpenSSL 3.5+, current browsers, OTP 28 `ssl`).

**Design**
- Uses the ML-KEM primitives OTP 28 exposes via `crypto` (`generate_key(mlkem768, ...)`, `encapsulate_key/2`, `decapsulate_key/3`) — no new dependencies.
- Wire format per the draft: client share = encapsulation key ‖ X25519 public (1216 B), server share = ciphertext ‖ X25519 public (1120 B), shared secret = ML-KEM secret ‖ X25519 secret (64 B). Verified against OTP ssl's implementation of the same group.
- The server side of a KEM exchange is an encapsulation rather than keygen+ECDH, so `quic_crypto` gains `server_key_exchange/2` covering both shapes; classical groups keep their exact previous behavior through it.
- Everything else rides the existing multi-group machinery — `supported_groups` negotiation, HelloRetryRequest into the hybrid group, key_share plumbing — with just the new code point added.
- Inert on crypto without ML-KEM (`quic_crypto:group_supported/1`); the new CT suite skips itself there.

**Testing**
- New `quic_pqc_e2e_SUITE`: direct hybrid negotiation, classical-client interop against a hybrid-preferring server, hybrid negotiated via HRR (exercises hybrid keygen on the CH2 retry), and wire-format/shared-secret invariants.
- Full existing test suite passes: 2247 eunit tests, `quic_hrr_e2e_SUITE`, TLS negotiation/compliance/server eunit modules — 0 failures.

**Why**: harvest-now-decrypt-later makes quantum-safe key exchange the PQC piece with a deadline; C-based QUIC stacks only get it via OpenSSL 3.5+. With this, a pure-Erlang QUIC endpoint offers it from any OTP 28 install, at ~1.2 kB extra ClientHello when the group is offered. Defaults are unchanged (`default_groups()` is still `[x25519]`) — flipping the default hybrid-first could be a follow-up decision.

Related: on the OTP side I have a PR open exposing the negotiated group via `ssl:connection_information` (erlang/otp#11440) for the same PQC-observability motivation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Post-quantum hybrid key exchange `x25519mlkem768` (draft-ietf-tls-ecdhe-mlkem, ML-KEM-768 + X25519), opt-in via `groups`. Negotiable only when the crypto library provides ML-KEM-768 (OTP 28+); a `groups` option naming an unsupported group is rejected up front from `connect/4` and `start_server/3` as `{error, {unsupported_group, _}}` instead of crashing during the handshake. (#195)
+
+### Fixed
+- Initial-level CRYPTO is now chunked across Initial packets, each within the 1200-byte pre-PMTU limit, and the whole flight is replayed on retransmit. A hybrid `x25519mlkem768` ClientHello (~1360 bytes) or ServerHello Initial (~1225 bytes) no longer leaves as a single oversized datagram that is dropped on paths with an MTU below ~1470 (IPv6-over-PPPoE, WireGuard, mobile). (#195)
+
 ## [1.8.0] - 2026-08-05
 
 ### Fixed
--- a/docs/CLIENT_GUIDE.md
+++ b/docs/CLIENT_GUIDE.md
@@ -45,7 +45,7 @@ quic:close(Conn, normal).
 | `server_name` | binary | Host | Server Name Indication |
 | `cert` | binary | - | Client certificate (for mTLS) |
 | `key` | term | - | Client private key (for mTLS) |
-| `groups` | [atom()] | `[x25519]` | Key-exchange groups in preference order (`x25519`, `secp256r1`, `secp384r1`). The head gets a `key_share`; the rest are HelloRetryRequest-eligible. |
+| `groups` | [atom()] | `[x25519]` | Key-exchange groups in preference order (`x25519`, `secp256r1`, `secp384r1`, `x25519mlkem768`). The head gets a `key_share`; the rest are HelloRetryRequest-eligible. `x25519mlkem768` is the post-quantum hybrid group (needs OTP 28+); naming an unsupported group returns `{error, {unsupported_group, _}}`. |
 | `signature_algs` | [atom()] | historical list | Advertised signature schemes (`ecdsa_secp256r1_sha256`, `ecdsa_secp384r1_sha384`, `rsa_pss_rsae_sha256\|384\|512`, `ed25519`, `rsa_pkcs1_sha256`). |
 | `external_psk` | tuple | - | TLS 1.3 external PSK; see [PSK.md](PSK.md). |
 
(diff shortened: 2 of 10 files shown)


benoitc:
Thanks, this is nice work. Suite passes on OTP 29, and I checked the wire format on my side too: 1216 / 1120 / 64 bytes, ML-KEM secret first. The `server_key_exchange/2` split is the right call, an encapsulation is not a keygen + ECDH.

Before merging, please fix the size on the wire:

With the hybrid group the ClientHello goes from 173 to 1359 bytes, and it leaves as a single datagram of 1445 bytes. I measured it with a plain UDP socket in place of the server. On the server side the ServerHello Initial is 1225 bytes (from its qlog). The Handshake flight is fine, it is already chunked.

The tests do not see it because loopback has a 16k MTU. On a real path 1445 + headers = 1473, so it passes on clean ethernet and not below: IPv6 over PPPoE is 1492 and already fails, WireGuard ~1420, mobile ~1400. And before the handshake there is no PMTU, so the safe size is 1200.

We already have what is needed here: `send_handshake_crypto/2` + `chunk_crypto/3` + `handshake_crypto_budget/1`, added in #134 for exactly the same problem one level up ("a single oversized datagram is dropped by strict clients"). The Initial path never needed it until now. So the ClientHello, the HRR retry in `handle_hello_retry_request/3` (same shape) and the ServerHello should all go through that chunking.

> Attention: `initial_crypto_frame` keeps a single frame for the retransmit, it will have to keep every chunk.

Small things:

- `rebar3 fmt --check` fails on `src/quic_crypto.erl`, the `compute_shared_secret/3` head is too long.
- no CHANGELOG entry.
- `group_supported/1` is exported but never called in `src/`. On OTP 27, someone who sets `groups => [x25519mlkem768]` gets a crash from `crypto:generate_key` instead of a clean error, better to check the option at connect/listen.
- `docs/CLIENT_GUIDE.md` documents `groups` as well, not only features.md.

I approved the CI run so you get the checks.

jbevemyr:
Thanks for the careful review — all addressed, pushed.

**The wire size (the important one).** You're right, and it was the Initial path specifically. The ClientHello, the ServerHello, and the CH2 built in `handle_hello_retry_request/3` now all go through a `send_initial_crypto/3` that chunks the flight the same way `send_handshake_crypto/2` does one level up (#134), with a budget that accounts for the Initial token field. Each Initial packet is separately padded to 1200, so a hybrid ClientHello leaves as two 1200-byte datagrams.

Good catch on the retransmit — `initial_crypto_frame` became `initial_crypto_frames`, the whole flight is stored and replayed. After an HRR the buffer is replaced with the CH2 chunks (the outstanding flight), which also fixes a latent staleness where a retransmit would have resent CH1.

New CT suite `quic_pqc_mtu_SUITE` measures it the way you did — drives a full hybrid handshake through a socket adapter and asserts every Initial datagram on the wire is ≤ 1200 and that the ClientHello spans more than one Initial. Captured sizes: `[1200,1200,1200,1200,1200]`.

**The small things:**
- `group_supported/1` is now wired in: `validate_groups/1` rejects an unsupported `groups` entry from both `connect/4` and `start_server/3` as `{error, {unsupported_group, _}}` — so an OTP 27 node setting `x25519mlkem768` gets a clean error, not a `crypto:generate_key` crash. Test added.
- `rebar3 fmt --check` is clean now (reflowed the `compute_shared_secret/3` head).
- CHANGELOG entry under `[Unreleased]`.
- `docs/CLIENT_GUIDE.md` documents the hybrid group, the auto-chunking, and the unsupported-group error.

Full suite green here (2247 eunit, HRR + PQC CT, fmt --check). Ping me if anything else.

jbevemyr:
The Elvis job on the last run was my fault: wiring in the groups validation nested a second `case` in `connect/4` and pushed it to 33 lines, over the `max_function_length` limit of 30. Composed the two client-side checks into `validate_client_opts/2` instead, so `connect/4` is back to its original shape. Pushed as 81bdfe7.

Verified locally before pushing this time: `rebar3 lint`, `fmt --check`, `dialyzer`, eunit (2247), and the PQC/MTU/HRR CT suites all exit 0. Everything else in that run was already green, including all four OTP versions.

The new run needs your approval to start (fork PR gate).

benoitc:
Superseded by #198, which carries your commits merged with main plus review fixes. Thanks!

PR #200, opened 2026-08-15
Derive the GSO segment size from the batch

### LINKED ISSUE #196: GSO never engages on the 1-RTT path (gso_size fixed at 1200, packets are 1398), and enabling it breaks the connection
While benchmarking the socket backend I found that the GSO send path never executes, and that making it execute breaks delivery. Two separate things, reporting both. No patch proposed — the second half depends on the receive path.

Environment: Linux 6.8, OTP 28.5.0.1, `socket_backend => socket` on client and listener, `quic_socket:detect_capabilities/0` reports `#{backend => socket, gso => true, gro => true}`.

## 1. GSO never fires

`flush/1` only takes the GSO path when `gso_batch_uniform(Buffer, GSO)` holds, which requires every packet except the last to be exactly `gso_size`. `gso_size` defaults to `?DEFAULT_GSO_SEGMENT_SIZE` (1200), but 1-RTT packets are sized from the current max datagram size and come out at **1398 bytes**.

Instrumented `gso_batch_uniform/2` over a 10 MB single-stream transfer, logging `{Result, GSO, FirstPacketSize}`:

```
{false, 1200, 1398} => 108
```

Never true, so every flush falls through to `flush_individual/1`. Counting flushes by batch size over the same transfer:

```
individual_flushes=3939  packets=6214  batch_size_1=3898
```

So ~1.6 packets per syscall, and `flush_gso/1` is dead code in practice. The batching machinery itself works — `add_to_batch/2` accumulates correctly up to `max_batch_packets`, and a minority of flushes do carry ~56 packets — it is only the uniformity comparison against the fixed 1200 that fails.

Related: the preemptive flush before an ACK-only packet in `quic_connection` (the comment referencing `gso_batch_uniform/2`) is protecting a path that never runs today.

## 2. Enabling it breaks the connection

I tried the obvious change locally: derive the segment size from the batch (all packets except the last must share a size, use that as `UDP_SEGMENT`) instead of comparing to `gso_size`, and split writes so one `sendmsg` stays under the 64 KB GSO payload cap.

It compiles clean and the GSO path then runs — but the connection no longer completes a transfer. Even 1 MB fails:

```
{closed, idle_timeout}
```

and larger transfers die mid-flight with `{error, {invalid_state, draining}}`. Reverting restores ~26 MB/s on loopback. I did not chase this further because the likely interaction is on the receive side (whether the `UDP_GRO` cmsg is delivered and `split_gro_packets/2` applied on the active-mode paths, and how loopback segmentation behaves), which you know far better than I do.

## Why it matters

Per-packet sends put a hard ceiling on throughput on any real path. For calibration, a raw `gen_udp` sender through a 20 ms `netem` qdisc on loopback delivers ~6,300 pps (~8.1 MB/s) end-to-end at 1398-byte datagrams; erlang_quic reaches ~4 MB/s there, i.e. it is close to the per-packet ceiling of the path rather than to what coalescing would allow. A C stack using GSO on the same path is not bound by that ceiling because the kernel sees a few large writes instead of thousands of small ones.

Happy to retest any change you make, or to hand over the instrumentation patches (batch-size histogram, uniformity probe, per-gate blocked counters) if useful — they were about 20 lines and I threw them away to keep the branch clean.

benoitc:
Fixes #196.

The GSO path never ran: the uniformity check compared every packet against a `gso_size` that no caller sets, so it stayed at the 1200 default while 1-RTT packets follow the current max datagram size. Deriving the size from the batch is the only value that can be right.

That alone is not enough to turn it on. `flush_gso/1` also read the segment size from the record rather than from the batch, so making the check pass still told the kernel to cut at 1200 in the middle of 1398-byte packets, which is the `idle_timeout` reported in the issue. The listener path also set a socket-level `UDP_SEGMENT`, which segments every datagram including handshake packets; GSO is now requested per message only, as `build_socket_state` already documented.

CI could not catch any of this because `batch_flushes` and `packets_coalesced` advance on the individual path too, so the download test's ratio measured batching rather than offload. `gso_flushes` counts only the batches the kernel segmented, and the Linux job now asserts it advanced.

@jbevemyr this is your report; if you can retest on your setup that would be welcome, since the maintainer's machine cannot exercise the path.

### DIFF
--- a/include/quic.hrl
+++ b/include/quic.hrl
@@ -402,8 +402,11 @@
 %% UDP Packet Batching (GSO/GRO on Linux)
 %% Maximum packets to batch before auto-flush
 -define(DEFAULT_MAX_BATCH_PACKETS, 64).
-%% Default GSO segment size (QUIC packet size for batching)
--define(DEFAULT_GSO_SEGMENT_SIZE, 1200).
+%% Kernel limits on a single UDP_SEGMENT write: UDP_MAX_SEGMENTS
+%% segments, and a payload that still fits a 16-bit UDP length. A batch
+%% larger than either is split across writes.
+-define(MAX_GSO_SEGMENTS, 64).
+-define(MAX_GSO_PAYLOAD, 65535).
 
 %%====================================================================
 %% Records
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -2122,7 +2122,7 @@ connected(
     %% Return packet counts for liveness detection (net_kernel uses
     %% recv count to verify peer is alive) plus send-path batching
     %% counters for benchmarks and tests.
-    {Flushes, Coalesced} = send_batch_counters(SocketState),
+    {Flushes, Coalesced, GSOFlushes} = send_batch_counters(SocketState),
     Stats = #{
         packets_received => PacketsRecv,
         packets_sent => PacketsSent,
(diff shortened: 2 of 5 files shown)


benoitc:

Issue #201, opened 2026-08-17
Small sends do not coalesce: one datagram per send_data, 10x throughput left on the table


jbevemyr:
We are evaluating erlang_quic as a production transport for a pub/sub system whose dominant traffic is many small messages multiplexed over one connection (hundreds of streams, messages of 50-500 bytes, bursts of tens of thousands). On that shape the send path currently pays one datagram (or more) per `send_data`, and it is the single largest performance gap we measured against our current C-based stack.

## Measurement

Loopback, one connection, one stream, 20 000 messages of 100 bytes, 1.7.0. UDP sends counted with call tracing on the client connection process only:

| mode | rate | `gen_udp:send` per message |
|---|---|---|
| `send_data/4` (sync) per message | 15.5k msg/s | 2.00 |
| `send_data_async/4` per message | 21.4k msg/s | 1.49 |
| 10 messages concatenated per `send_data` | **154k msg/s** | 0.20 |

The 10x jump from application-side batching is the tell: the cost is per-send packetization, not bytes. Every `send_data` becomes its own STREAM frame in its own packet in its own datagram, so a 100-byte message ships with ~60 bytes of packet overhead plus a syscall, and the ACK-eliciting packet rate explodes (the 2.00 includes ACK traffic provoked at the peer).

For comparison, msquic buffers app sends and fills packets to the MTU: the same message stream coalesces into ~10-14 messages per datagram.

## Proposal: fill packets from queued small frames at packetization time

The machinery is nearly there already. `dequeue_small_stream_frame_tuple/1` opportunistically piggybacks one queued small stream frame onto an outgoing ACK packet. Generalizing that:

1. When building a 1-RTT packet for a stream frame, after placing the triggering frame, keep draining further queued small stream frames (any stream, respecting per-stream and connection flow control and cc) into the same packet until the datagram budget is spent - the classic packet-fill loop.
2. Route small direct sends through the send queue instead of packetizing immediately when the queue is non-empty, so consecutive `send_data_async` casts that are already sitting in the connection process mailbox get the chance to share datagrams. A pure mailbox-drain approach (process all pending `{send_data,...}` casts before packetizing) would achieve the same without any timer.
3. No aggregation delay needed: only coalesce what is already there. Latency for a lone small send stays one packetization, unchanged.

[...]

PR #203, opened 2026-08-17
Coalesce small sends into shared packets (#201)

jbevemyr:
Implements the proposal from #201: coalesce small sends into shared packets.

## What it does

`send_data`/`send_data_async` handlers drain consecutive send requests already queued in the connection mailbox (bounded at 64), process them as one batch, and accumulate the resulting frames into shared packets up to the PMTU budget via a pending-packet buffer in front of `send_app_packet_internal`; the batch flushes as one (or few) multi-frame packets. Everything downstream - packet build, loss tracking, retransmission, counters - already handles a frames list per packet, so only the emission point changes.

- A lone send behaves exactly as before, apart from one empty `receive ... after 0` probe.
- Ordering within a batch is arrival order; only send messages are drained, so other requests keep their ordinary ordering relative to each other.
- Coalescing is only active inside one handler invocation; no timers, no added latency for isolated sends.

## Numbers

Loopback, 20k x 100-byte messages on one connection (the workload from #201):

| scenario | before | after |
|---|---|---|
| 8 concurrent senders (sync) | 15.5k msg/s, 2.0 datagrams/msg | **66.9k msg/s, 0.35 datagrams/msg** |
| single sender, sync | 15.5k msg/s | unchanged (cannot batch with itself) |

In our production-shaped integration benchmark (message pipeline over one connection, mixed streams), receiver-side CPU dropped 17% and sender-side wall time 9% from this change alone, with a further step when the application switched to `send_data_async` (its casts queue and batch even from a single producer): sender CPU gap versus our C-stack baseline shrank from +71% to +24%.

## Notes

- Tests: `rebar3 eunit` identical pass/fail set to `main` before and after (the 16 failures on my machine are environmental cert-generation issues, present on clean main).
- One pre-existing oddity noticed while working here: the flow-control-blocked paths in `do_send_data` build and send DATA_BLOCKED / STREAM_DATA_BLOCKED via `send_frame` but discard the returned state (`_FinalState`). Under coalescing the frame ends up in a discarded pending buffer instead (net effect: advisory frame dropped rather than sent with a possibly-reused packet number). Left as-is; may deserve its own fix.

Same setup as #196/#201/#202; happy to adjust.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -685,7 +685,18 @@
     last_recv_trigger = sequential :: sequential | reordered,
 
     %% QLOG Tracing (draft-ietf-quic-qlog-quic-events)
-    qlog_ctx :: #qlog_ctx{} | undefined
+    qlog_ctx :: #qlog_ctx{} | undefined,
+
+    %% Small-send coalescing (issue #201). While `coalesce' is true -
+    %% only within one drained batch of send_data/send_data_async
+    %% requests - app packets accumulate here and are flushed as one
+    %% multi-frame packet per PMTU budget instead of one packet per
+    %% frame. Reversed accumulation lists; see send_app_packet_internal
+    %% and flush_pending_packet.
+    coalesce = false :: boolean(),
+    pend_payload = [] :: [iodata()],
+    pend_frames = [] :: [tuple()],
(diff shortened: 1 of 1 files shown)


benoitc:

benoitc:

PR #204, opened 2026-08-17
GRO: accept the int-sized UDP_GRO cmsg payload

jbevemyr:
The one-line fix from the #196 retest, as its own PR: `extract_gro_segment_size/1` matched exactly two bytes of cmsg data, but the kernel delivers the segment size as a native int, so the match always failed and GRO-coalesced buffers were passed up unsplit (and dropped - a short-header packet consumes the rest of the datagram, so merged 1-RTT flights cannot be recovered). Verified end to end in the #196 setup: the previously 0/7-failing netem transfer path went 8/8 with this fix on top of #200. Independent of #200 (the parse bug exists regardless), but only observable once a peer's GSO actually engages.

### DIFF
--- a/src/quic_socket.erl
+++ b/src/quic_socket.erl
@@ -1208,7 +1208,14 @@ recv_gro(Socket, Timeout) ->
 
 extract_gro_segment_size([]) ->
     undefined;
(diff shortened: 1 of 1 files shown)


benoitc:

benoitc:

PR #205, opened 2026-08-18
Fix async-send data loss and a send-queue ordering deadlock

jbevemyr:
Two related data-loss/deadlock bugs in the send path, found while running this library as the QUIC client against an msquic server under sustained bulk load with `{return, async}`-style fire-and-forget sends.

## 1. Async sends silently dropped (`d2bdd4b`)

`send_data_async/4` is fire-and-forget, but two paths returned errors that no caller can see:

- **Flow-control blocked** (connection or stream level): `do_send_data` returned `{error, {flow_control_blocked, _}}` and dropped the payload. The stream's `send_offset` advances on the next send, so the peer sees a clean, contiguous stream with a chunk of application data simply missing — silent corruption rather than a visible failure. Fixed by queueing the data on the send queue (the DATA_BLOCKED / STREAM_DATA_BLOCKED frame is still emitted) and letting `process_send_queue/1` retry when MAX_DATA / MAX_STREAM_DATA arrives. `send_offset` is advanced at queue time so later sends order behind the queued entry; reassembly is offset-based, so cross-entry transmission order stays correct.

- **Cast arriving before `connected`**: the owner learns `connected` a flight before the server's confirmation transitions the state machine, so an immediate `send_data_async` could hit `handshaking` and fall into the catch-all, which dropped it. Queued into `pending_data` like the synchronous handshaking-state clause (flushed on entering `connected`).

## 2. Requeue ordering deadlock (`caa5dd5`)

When a send-queue drain pops an entry and congestion control or pacing blocks mid-entry, the unsent remainder was re-queued at the **back** of its urgency bucket — behind higher-offset entries of the same stream. The drain then transmits those higher offsets, the receiver buffers them out of order behind the hole, and once the queue head reaches the stream flow-control limit the drain stops permanently: the peer cannot extend a window that tracks contiguous delivery (stuck at the hole), and the only entry that could fill the hole is unreachable behind the blocked head. The connection deadlocks with the sender showing `in_flight = 0` and everything acked.

We caught this on the wire against msquic: a single entry-sized hole (e.g. `[133426,137580)`) never transmitted, ~126 KB beyond it delivered and buffered, and the peer's frozen `MAX_STREAM_DATA` grant equal to hole base + receive window exactly (264498 = 133426 + 131072).

Fix: a `pqueue_in_front/3` primitive; drain remainders are front-inserted, preserving per-stream offset order (fresh sends keep appending at the back — their offset is always highest). Lost-frame retransmits (`enqueue_retransmit_stream`) are front-inserted for the same reason: they exist to fill receiver-side holes, so they must not queue behind a flow-control-blocked head whose window can only grow once the hole is filled.

## Validation

- `rebar3 eunit` 2263/2263, `dialyzer`, `xref`, `lint`, `fmt --check` all clean.
- Application-level soak (volga event streaming, this library as client vs msquic 2.4 server, multi-MB topic replication): the bulk-transfer stall reproduced ~100% before, 5/5 clean runs after.

Independent of #203/#204 (applies directly to main); the fixes also compose with the coalescing branch.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -8476,7 +8476,15 @@ queue_stream_data(
     } = State,
     Where
 ) ->
-    DataSize = iolist_size(Data),
+    %% Normalise to a binary: consumers like
+    %% dequeue_small_stream_frame_tuple/1 hand the entry's data straight
+    %% to quic_frame:encode/1, which requires a binary.
+    DataBin =
+        case is_binary(Data) of
+            true -> Data;
+            false -> iolist_to_binary(Data)
+        end,
+    DataSize = byte_size(DataBin),
     NewQueueBytes = QueueBytes + DataSize,
     case NewQueueBytes > ?MAX_SEND_QUEUE_BYTES of
         true ->
@@ -8494,7 +8502,7 @@ queue_stream_data(
         false ->
             Urgency = get_stream_urgency(StreamId, Streams),
             %% Cache DataSize in entry to avoid repeated iolist_size calls
-            Entry = {stream_data, StreamId, Offset, Data, Fin, DataSize},
+            Entry = {stream_data, StreamId, Offset, DataBin, Fin, DataSize},
             NewPQ =
                 case Where of
                     back -> pqueue_in(Entry, Urgency, PQ);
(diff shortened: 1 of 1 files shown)


jbevemyr:
Pushed one more commit (66ce721): queued stream data is now normalised to a binary at queue time. The flow-control-blocked queueing added in this PR stored the caller's iodata as-is, but `dequeue_small_stream_frame_tuple/1` hands small queue entries straight to `quic_frame:encode/1`, which requires a binary — an iodata send (e.g. `[<<Len:32>>, Payload]`) blocked on flow control crashed the connection with badarg in `byte_size/1` when a later packet picked it up. Caught in application soak testing; eunit and dialyzer clean.

benoitc:
Duplicate of #225, which reworks the same path more completely. Closing in favour of it.

PR #206, opened 2026-08-18
Handle overlapping stream frames in the receive path

jbevemyr:
RFC 9000 §2.2 allows a sender to repacketize stream data on retransmission, so an incoming STREAM frame can overlap data the receiver has already delivered or buffered. The receive path assumed exact framing in two places:

- A frame starting below `recv_offset` was buffered at its original offset, where `extract_contiguous_data` (exact-offset `maps:take`) never visits it.
- Extraction could not cross a buffered frame that covers the contiguous point without starting exactly on it.

Since the packets carrying the overlapped bytes were acked, the sender never resends them: the stream stalls permanently with the tail parked in the out-of-order buffer. We hit this reliably against an msquic sender under a bursty ~5 MB backlog dump on loopback (UDP loss + msquic repacketization): consumer frozen mid-transfer with ~9 MB parked in `recv_buffer`, `recv_offset` stuck at a hole no incoming frame starts at, and stale below-`recv_offset` buffer entries proving prior overlapping deliveries.

Fix:
- `buffer_stream_data/4`: trim the already-delivered head before buffering; on an exact offset collision keep the longer payload.
- `extract_contiguous_data/2`: when no frame starts exactly at the contiguous point, fall back to the buffered frame reaching furthest past it and deliver its unseen tail.
- Prune buffered entries wholly below the delivered point when extraction stops at a genuine gap.
- Duplicate classification and connection-level `data_received` accounting now count only bytes past the delivered point.

Five new eunit tests cover head-trim, full-duplicate, same-offset-longer-wins, covering-frame extraction, and stale-entry pruning. `rebar3 eunit` 2268/2268, `dialyzer`, `fmt --check` clean. Application-level validation: the stalling integration test went from failing ~1 in 3 to passing repeatedly with this fix.

Independent of #203/#204/#205 (applies directly to main).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -140,6 +140,8 @@
 -ifdef(TEST).
 -export([
     chunk_crypto/3,
+    buffer_stream_data/4,
+    extract_contiguous_data/2,
     add_to_ack_ranges/2,
     merge_ack_ranges/1,
     convert_ack_ranges_for_encode/1,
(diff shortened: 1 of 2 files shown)


benoitc:
Duplicate of #223, which reworks the same path more completely. Closing in favour of it.

PR #207, opened 2026-08-19
Hibernate idle connection processes

jbevemyr:
An idle gen_statem never garbage-collects on its own, so every connection keeps the garbage produced by its handshake pinned to the process heap indefinitely. Measured with 300 idle established connections per side (loopback, in-VM echo server): **~177 KiB per server connection** (~164 KiB client) sitting idle after the handshake, against ~7 KiB of live state. On a server holding tens of thousands of mostly-idle connections this dominates the VM footprint — roughly 9 GB of dead heap at 50k connections.

This starts the connection statem with `{hibernate_after, T}`: a connection with no event for T ms hibernates (fullsweep GC, stack released, heap compacted to live state). Measured effect on the same bench:

| | before | after |
|---|---|---|
| avg server conn process | 177.4 KiB | **7.7 KiB** (23x) |
| avg client conn process | 164.3 KiB | 7.0 KiB |
| VM growth, 600 idle conns | 117 MB | 12.5 MB |

Busy connections never reach the timer, so throughput is unaffected; a hibernated connection wakes on any event at microsecond cost. T comes from a new `hibernate_after` connection/server option (default 5000 ms, `infinity` disables).

`rebar3 eunit` 2263/2263, `dialyzer`, `fmt --check` clean; also validated under our application integration suites (bulk transfer and API tests green with the change).

Possible follow-up (not in this PR): the remaining ~7 KiB of live state still carries handshake-era fields — `initial_keys`/`handshake_keys` (RFC 9001 §4.9 says to discard these after the handshake), the CRYPTO reassembly buffer, and the Initial/Handshake packet-number spaces. Dropping them post-handshake would shave another ~2-3 KiB per connection and improve spec hygiene.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -711,7 +711,7 @@ start_link(Host, Port, Opts, Owner) ->
     gen_udp:socket() | undefined
 ) -> {ok, pid()} | {error, term()}.
 start_link(Host, Port, Opts, Owner, Socket) ->
-    gen_statem:start_link(?MODULE, [Host, Port, Opts, Owner, Socket], []).
+    gen_statem:start_link(?MODULE, [Host, Port, Opts, Owner, Socket], statem_opts(Opts)).
 
 %% @doc Initiate a connection to a QUIC server.
 %% This is a convenience wrapper that starts the process and initiates handshake.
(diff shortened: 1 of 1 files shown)


benoitc:

PR #208, opened 2026-08-19
Give up on an unresponsive peer with a disconnect timeout

### LINKED ISSUE #202: Client holds a dead connection for the full idle timeout when the peer restarts: no PTO-based disconnect
Follow-up from the same production evaluation as #201, this time on failure detection.

## Scenario

A client holds a long-lived connection to a server whose *process* restarts (the listener comes back on the same address within seconds). Captured with erlang_quic 1.7.0 as client against an msquic 2.4 server:

1. After the restart the client's packets hit a server with no connection state; it replies with stateless resets (observed as a stream of 30-byte short-header datagrams the client logs as `[short header decryption failed]`, one every few seconds, indefinitely).
2. The client cannot recognize them as resets: RFC 9000 section 10.3 recognition is exact-token, and msquic derives reset tokens from a per-process key, so the restarted process advertises tokens the surviving client never learned. This is not fixable on the client side by token matching, and any peer implementation with non-stable reset keys behaves the same after a restart.
3. Idle handling is correct (activity only bumps on successfully processed packets - verified), so the connection does die eventually. But with keep-alives configured, the natural `idle_timeout` is a multiple of the keep-alive interval: in our configuration 60 s of zombie connection during which every `send_data` is silently accepted into a black hole.

For comparison, msquic's client abandons the same dead path in ~16 s via its disconnect timeout: repeated PTO expiries with no acknowledgment are treated as path failure, independent of the idle timeout.

## Proposal: give up after persistent PTO expiry

RFC 9002 section 6.2 deliberately leaves this open ("endpoints could implement a maximum PTO count or time threshold"). Concretely:

- a `max_pto_backoff` (or `disconnect_timeout`) connection option: when ack-eliciting data has been outstanding continuously for that long - i.e. the PTO timer has fired and doubled N times with no packet successfully processed in between - close with a transport error to the owner (`{closed, pto_exhausted}` or similar);
- default it generously (say 30 s) or leave it opt-in to preserve current behavior;
- the state needed is already present (loss recovery tracks PTO count; the reset happens on any successfully processed packet).

This bounds the zombie window for *every* undetectable-death case (process restart, NAT box reboot, silent path loss with an active sender), not just the stateless-reset one, and matches what deployed stacks do.

The heuristic alternative - treating N consecutive short-header decryption failures from the peer's address as probable reset - would also work for this specific case, but PTO exhaustion is the more principled and general signal, and cannot be triggered by an off-path attacker injecting garbage.

Reproduction: any request/response or streaming workload where the server process restarts mid-stream while the client has data outstanding; the client stalls until idle timeout instead of failing fast. Happy to validate a patch against our integration suite where we caught this.

jbevemyr:
Fixes #202.

A connection whose peer stops responding while ack-eliciting data is in flight previously lived until the idle timeout, sending PTO probes into the void the whole time. The common way to get such a peer is a restart: RFC 9000 stateless resets require the exact token, and peers that derive reset tokens from per-process state (msquic does) send resets the surviving side can never recognise - so it correctly ignores them and hangs for the full idle timeout, tens of seconds of blackout for the application. msquic escapes the same situation with its 16 s disconnect timeout.

This adds the same mechanism:

- When a PTO fires with data in flight and the last forward progress is more than `disconnect_timeout` ms ago, the connection closes and the owner is notified with reason `disconnect_timeout`.
- Forward progress = the last received ACK **or** the start of the current outstanding burst (tracked in `quic_loss` as `outstanding_since`, set when an ack-eliciting packet is sent while nothing was in flight) - so a fresh send after a long quiet period cannot trip the timeout with a stale ACK timestamp.
- New connection option `disconnect_timeout`, default 16000 ms, `infinity` disables and restores the old behaviour. Idle connections are unaffected: with nothing in flight the check never applies.

Includes an e2e test that freezes the server-side connection process and asserts the client closes with reason `disconnect_timeout` in ~3 s (2 s timeout + PTO cadence) instead of waiting out a 60 s idle timeout.

`rebar3 eunit` 2263/2263, `ct --suite=quic_server_e2e_SUITE` 5/5, `dialyzer`, `fmt --check` clean.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -491,6 +491,13 @@
     keep_alive_interval :: non_neg_integer() | disabled,
     keep_alive_timer :: reference() | undefined,
 
+    %% Give up on an unresponsive peer: close the connection when
+    %% ack-eliciting data has been in flight with no ACK for this long
+    %% (RFC 9000 permits idle-timeout-only, but a stateless-reset-blind
+    %% peer then lingers for the whole idle timeout; msquic uses a 16 s
+    %% disconnect timeout for the same reason).
+    disconnect_timeout = 16000 :: pos_integer() | infinity,
+
     %% Pacing (RFC 9002 Section 7.7)
     pacing_timer :: reference() | undefined,
     pacing_enabled = true :: boolean(),
@@ -1172,6 +1179,7 @@ init({server, Opts}) ->
         reset_stream_at_enabled = maps:get(reset_stream_at, Opts, false),
         idle_timeout = IdleTimeout,
         keep_alive_interval = calculate_keep_alive_interval(Opts, IdleTimeout),
+        disconnect_timeout = disconnect_timeout_opt(Opts),
         keep_alive_timer = undefined,
         last_activity = erlang:monotonic_time(millisecond),
         cc_state = CCState,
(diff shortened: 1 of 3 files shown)


benoitc:
Duplicate of #224, which reworks the same path more completely. Closing in favour of it.

PR #210, opened 2026-08-19
Use max(smoothed_rtt, latest_rtt) for the loss time threshold (RFC 9002 §6.1.2)

jbevemyr:
RFC 9002 §6.1.2 defines the loss time threshold as `kTimeThreshold * max(smoothed_rtt, latest_rtt)`; the code used `smoothed_rtt` alone.

The smoothed EWMA under-estimates the threshold whenever the path RTT rises faster than the EWMA follows (receiver queueing, bufferbloat). Every in-flight packet older than ~1.125× the stale smoothed RTT then gets declared lost even though its ACK is merely late. Each spurious loss retransmits already-delivered data and collapses the congestion window — exactly the wrong response to queueing delay.

Observed on a loopback bulk transfer with an overloaded receiver: thousands of spurious retransmits per 200 MiB, a large fraction attributable to the time threshold firing during RTT spikes (verified by instrumenting which threshold declared each loss).

Applied at all three sites that compute the threshold: `on_ack_received/3` phase-2 detection, `detect_lost_packets/2`, and `get_loss_time_and_space/1` (timer arming).

### DIFF
--- a/src/quic_loss.erl
+++ b/src/quic_loss.erl
@@ -239,7 +239,7 @@ on_ack_received(State, {ack, LargestAcked, AckDelay, FirstRange, AckRanges}, Now
             {LostList, SurvHeadQ, LostBytes, LargestLostSentTime} =
                 detect_lost_q(
                     KeptList,
-                    NewState1#loss_state.smoothed_rtt,
+                    max(NewState1#loss_state.smoothed_rtt, NewState1#loss_state.latest_rtt),
                     LargestAcked,
                     Now,
                     [],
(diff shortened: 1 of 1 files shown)


benoitc:
Carried into #260 with tests, authorship intact.

Worth saying what changed my mind about this one. I first wrote it up as an RFC conformance fix, because the description argues from RFC 9002 6.1.2 and I could not see a failure. It is more than that: the smaller threshold declares real in-flight packets lost while their ACKs are merely late.

Three packets, two sent 20 ms before the third, then only the third acknowledged with a 400 ms sample. The earlier two are 420 ms old, past 1.125 * smoothed_rtt (75 ms) but inside 1.125 * latest_rtt (450 ms). On main they come back in the lost list; with your change they do not.

Two attempts at that test were wrong before it was right, both worth flagging since they would have made this look unfixed:

- the first spread the packets too far apart, so their age exceeded even the spiked threshold and they were lost either way
- the first also used six packets and acknowledged the highest, so the packet-number threshold (6.1.1, kPacketThreshold 3) declared the older ones lost regardless of time

Both are kept out of the final version: two packets, both within 3 of the acked packet number, aged to sit between the two thresholds. The time threshold is then the only thing deciding.

Will close this once #260 lands.

benoitc:
Duplicate of #260, which carries this work with tests. Closing in favour of it.

PR #211, opened 2026-08-19
Cap retained ACK ranges per PN space (RFC 9000 §13.2.4)

jbevemyr:
The per-PN-space `ack_ranges` list was unbounded. Under burst loss it fragments into hundreds of disjoint ranges, and since every outgoing ACK encodes the full list (and the peer decodes it), ACK processing cost grows O(ranges) per packet on **both** endpoints.

Profiled on a lossy loopback bulk transfer: ~270 ranges per ACK frame, kilobyte-sized ACK frames, and **57% of all connection-process CPU** spent in `encode_ack_ranges`/`decode_ack_ranges`/`ack_ranges_to_range_list` (6.6M calls for a 50 MiB transfer). The oversized ACKs also amplify the very congestion that caused the loss — a death spiral.

RFC 9000 §13.2.4 explicitly permits limiting retained ranges. This keeps the 64 highest ranges; packets below the lowest retained range are retransmitted by the peer if truly unacked, and discarded as duplicates otherwise. 64 is in line with other stacks (msquic caps its range tracking similarly).

The cap is applied at insertion (`update_pn_space_recv/3`), so encoding, decoding, and storage are all bounded by it.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -234,6 +234,15 @@
 %% ACK traffic for RTT-sample granularity.
 -define(ACK_PACKET_TOLERANCE, 2).
 
+%% Max ACK ranges retained per PN space (RFC 9000 §13.2.4 allows the
+%% receiver to limit these). Under burst loss an unbounded list
+%% fragments into hundreds of ranges, and since every outgoing ACK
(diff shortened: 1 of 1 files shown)


benoitc:

PR #212, opened 2026-08-19
Subject loss retransmissions to the congestion window

jbevemyr:
Loss retransmissions were sent via `can_send_control`, which allows exceeding the congestion window. RFC 9002 treats retransmitted data like any other send — it counts against cwnd; the overage allowance is meant for PTO probes (§7.5) and small control frames.

The practical failure mode: under sustained receiver overload, every loss epoch immediately refills the path with retransmit traffic that ignores the multiplicative decrease, so the drop/retransmit spiral never converges (observed as a large multiple of spurious+repeat retransmits per real drop on an overloaded loopback receiver).

The existing CC-deferral machinery (`retransmit_stream` queue entries and `deferred_ctrl_retransmits`) already handles the blocked case, so a blocked retransmission is deferred and replayed when cwnd reopens rather than dropped. PTO probes keep the control-allowance exemption via an explicit `probe` mode.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -9374,14 +9374,25 @@ filter_reset_stream_at_data(Frames, #state{streams = Streams}) ->
 %% Send frames for retransmission with congestion control check
 send_retransmit_frames_cc([], State) ->
     State;
-send_retransmit_frames_cc(Frames, #state{cc_state = CCState, retransmits = R} = State) ->
+send_retransmit_frames_cc(Frames, State) ->
+    send_retransmit_frames_cc(Frames, State, normal).
+
+%% Mode `probe' (PTO) may exceed the congestion window per RFC 9002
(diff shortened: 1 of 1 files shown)


benoitc:
See #249 for the detail: this and #249 make the same arity-2 to arity-3 refactor of `send_retransmit_frames_cc` and reach opposite conclusions, so they conflict and only one can land.

Short version. RFC 9002 section 7 exempts a PTO probe from cwnd and leaves an ordinary loss retransmission bound by it. This PR gets `normal -> can_send` right; #249 gets `probe -> true` right. main is wrong in both directions. The union of the two is the correct policy.

Also worth noting: I tested both against the window-stranding stall in #256 and neither changes the outcome, so the case here is conformance rather than a fixed defect. #250 is what fixes that stall.

jbevemyr:
Closing in favour of #249, which now carries the union of the two: probe exempt outright, normal bound by can_send. See the discussion there.

PR #213, opened 2026-08-19
Make 1-RTT ACK decimation tolerance configurable

jbevemyr:
Adds an `ack_packet_tolerance` connection option (default 2, the RFC 9000 §13.2.1 recommendation — behavior unchanged by default) controlling how many ack-eliciting 1-RTT packets are accumulated before an ACK is flushed. The `max_ack_delay` timer still bounds ACK latency regardless of the tolerance.

Motivation: on bulk transfers the sender pays a full decrypt + frame pass for every incoming ACK. At the default every-2nd-packet cadence, a 200 MiB single-stream transfer makes the sender process ~78k ACK packets — a significant share of its CPU. With tolerance 25 the same transfer delivers ~6.5k ACKs and runs ~35% faster (29 → 40 MiB/s, loopback), at the cost of coarser RTT samples and slower loss feedback — a trade-off the application should own, hence an option rather than a new default.

Also gates the per-acked-packet PMTU probe fold on a probe actually being outstanding; bulk ACK processing previously called `handle_pmtu_probe_ack` once per acked packet just to discover there was no probe.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -676,6 +676,12 @@
     %% armed so the peer sees an ACK at worst max_ack_delay ms after
     %% the first ack-eliciting packet in the window.
     ack_elicited_count = 0 :: non_neg_integer(),
+    %% How many ack-eliciting 1-RTT packets to accumulate before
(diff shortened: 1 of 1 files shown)


benoitc:

PR #214, opened 2026-08-19
Bound send bursts with a per-drain packet budget

jbevemyr:
A single send event could emit an unbounded run of packets: with a large cwnd and negligible pacing delay (low-RTT paths), one big send plus the queue drain produced hundreds of back-to-back packets with no ACK or loss feedback processed in between. Receivers that drain slower than the sender emits lose large chunks of every burst, congestion feedback arrives only after the whole burst is on the wire, and throughput collapses under retransmit storms (observed ~27% packet retransmit rate against a default-rmem loopback receiver with GSO batching).

This caps packets emitted per drain at `max_burst_packets` (option, default 64). When the budget is spent, the remainder is queued and a zero-delay pacing continuation is armed, so the connection returns to the event loop, processes pending ACK/loss feedback, and resumes with a fresh budget. The pacing timer/message is reused for the continuation so stale-reference handling and the drain path stay shared.

Measured (loopback bulk): retransmits on a GSO-batched send path drop ~3.5x; the plain gen_udp path stays within ~4% of its previous throughput.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -234,6 +234,10 @@
 %% ACK traffic for RTT-sample granularity.
 -define(ACK_PACKET_TOLERANCE, 2).
 
+%% Default per-drain send burst budget in packets (max_burst_packets
+%% option). See the burst_budget field.
+-define(DEFAULT_MAX_BURST_PACKETS, 64).
+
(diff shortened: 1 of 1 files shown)


benoitc:

PR #216, opened 2026-08-19
Optional AEAD-context NIF with pure-Erlang fallback

jbevemyr:
QUIC seals/opens every packet as its own AEAD unit, and OTP's `crypto:crypto_one_time_aead/7` re-runs the AES key schedule on every call — on bulk transfers the two per-packet crypto calls (payload AEAD + header-protection block) are the largest single CPU consumer in the connection process (~20% profiled).

This adds a small, strictly optional accelerator:

* `c_src/quic_crypto_nif.c` (~300 lines): keeps an OpenSSL `EVP_CIPHER_CTX` per key as a NIF resource, so the key schedule runs once per key and each packet only resets the nonce. Seal/open for aes_128_gcm / aes_256_gcm / chacha20_poly1305, plus AES-ECB blocks for header protection. Data arguments accept iodata (the hot send path passes `[Header | Chunk]` iolists). Contexts are per-process (connection processes are the only callers, and their NIF calls are sequential).
* `quic_aead_ctx.erl`: dispatcher with a per-process context cache that transparently falls back to the OTP crypto one-shot API when the NIF is not built or fails to load. `quic_aead`'s call sites route through it; **semantics are unchanged and the library remains fully pure-Erlang-capable** — no NIF is required to build, load, or run.
* Build is a best-effort rebar `pre_hook` (silently skipped without a C toolchain). The Makefile links the same libcrypto that OTP's own crypto NIF resolved at runtime (found via `ldd`), so loading this NIF can never pin an incompatible OpenSSL into the VM before `crypto.so` loads — that failure mode is real when OTP is built against a non-system OpenSSL.

Correctness: differential-tested byte-exact against `crypto:crypto_one_time_aead` / `crypto:crypto_one_time` for all three ciphers, including tag-failure and context-reuse-after-failure cases; the full eunit suite passes with the NIF active and with it absent.

Measured on a loopback bulk transfer: connection-process crypto share drops from ~20% to ~7% with per-packet calls (further gains come from batching entry points, kept out of this PR to keep it reviewable).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ certs/*.pem
 certs/*.key
 doc/
 rebar3.crashdump
+/priv/quic_crypto_nif.so
--- a/c_src/Makefile
+++ b/c_src/Makefile
@@ -3,6 +3,12 @@
 # loading this NIF never pins an incompatible OpenSSL into the VM
 # before crypto.so loads. Falls back to plain -lcrypto when the
 # lookup fails.
+#
+# Best effort by design: the NIF is a pure accelerator with a
+# pure-Erlang fallback, so a missing toolchain or missing OpenSSL
+# headers must not fail the build of a project that embeds this
+# library (erlang.mk runs this Makefile unconditionally when it
+# exists). A failed compile prints a warning and succeeds.
 
 ERTS_INCLUDE := $(shell erl -noshell -eval 'io:format("~ts/erts-~ts/include", [code:root_dir(), erlang:system_info(version)]), halt().')
 CRYPTO_SO := $(shell erl -noshell -eval 'io:format("~ts", [filename:join([code:lib_dir(crypto), "priv", "lib", "crypto.so"])]), halt().')
@@ -18,11 +24,18 @@ LDFLAGS += -L$(OPENSSL_LIBDIR) -Wl,-rpath,$(OPENSSL_LIBDIR)
 CFLAGS += -I$(OPENSSL_LIBDIR)../include
 endif
 
+all: $(PRIV)/quic_crypto_nif.so
+
 $(PRIV)/quic_crypto_nif.so: quic_crypto_nif.c
-	mkdir -p $(PRIV)
-	$(CC) $(CFLAGS) -fPIC -shared -o $@ $< -I"$(ERTS_INCLUDE)" $(LDFLAGS) $(LDLIBS)
+	@mkdir -p $(PRIV)
+	@if $(CC) $(CFLAGS) -fPIC -shared -o $@ $< -I"$(ERTS_INCLUDE)" $(LDFLAGS) $(LDLIBS); then \
+	    echo "quic_crypto_nif: built $@"; \
+	else \
+	    echo "quic_crypto_nif: build failed (missing toolchain or OpenSSL headers?);" \
+	         "the pure-Erlang crypto path will be used" >&2; \
+	fi
 
 clean:
 	rm -f $(PRIV)/quic_crypto_nif.so
 
-.PHONY: clean
+.PHONY: all clean
--- a/c_src/Makefile
+++ b/c_src/Makefile
@@ -1,0 +1,28 @@
+# Builds the optional AEAD-context NIF. Links against the SAME
+# libcrypto that OTP's own crypto NIF uses (resolved via ldd), so
+# loading this NIF never pins an incompatible OpenSSL into the VM
+# before crypto.so loads. Falls back to plain -lcrypto when the
+# lookup fails.
+
+ERTS_INCLUDE := $(shell erl -noshell -eval 'io:format("~ts/erts-~ts/include", [code:root_dir(), erlang:system_info(version)]), halt().')
+CRYPTO_SO := $(shell erl -noshell -eval 'io:format("~ts", [filename:join([code:lib_dir(crypto), "priv", "lib", "crypto.so"])]), halt().')
+OPENSSL_LIB := $(shell ldd "$(CRYPTO_SO)" 2>/dev/null | awk '/libcrypto/ && $$3 != "" {print $$3}')
+OPENSSL_LIBDIR := $(dir $(OPENSSL_LIB))
+
+CFLAGS ?= -O2 -Wall -Wextra
+LDLIBS := -lcrypto
+PRIV := ../priv
+
+ifneq ($(strip $(OPENSSL_LIBDIR)),)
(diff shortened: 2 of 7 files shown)


jbevemyr:
Added a runtime opt-out: QUIC_DISABLE_CRYPTO_NIF=1 (or QUIC_DISABLE_NIFS=1) skips loading the NIF at module load and leaves the pure-Erlang fallback in place. We use it for benchmarking the fallback and for fault isolation. Full eunit passes both with and without the variable set.

benoitc:
Thanks for this. I profiled it before merging: the AEAD is ~9% of connection-process CPU and the NIF's realistic ceiling is ~2%, half of which we get from reusing an ECB context via crypto_init/crypto_update with no C at all, so I'm taking the pure-Erlang half and leaving the send syscall (17.8%) as the real target.

jbevemyr:
Updated 2026-09-03: every row now points at a PR. The whole stack is #274 → #282 → #283 → #285 (each on top of the previous), plus #284 for the interop runner. In total we managed to increase performance for our use case like this:

## erlang_quic performance work: what it is, where it lives, what it measured

All throughput numbers are delivery-verified (byte count plus CRC checked at the sink) unless marked otherwise; single-stream bulk over loopback, socket backend, 200 MiB to 2 GiB transfers. Earlier figures we published in #220 (805 MiB/s and the msquic comparison) came from a harness that silently dropped data and were retracted there; nothing below relies on them. CT figures are CPU seconds of the server node in our 50-site production test on the same host, eq versus quicer/msquic doing the same job.

### Bulk throughput, in the order the fixes landed

| Step | Change | Measured (before to after) | Upstream |
|---|---|---|---|
| 0 | Baseline erlang_quic 1.8.x | 8 to 16 MiB/s | |
| 1 | MAX_DATA trigger, loss threshold, ACK-range cap, cwnd-subject retransmits, ack tolerance, burst budget, GRO truncation fix | 8 to 16, then ~30 gen_udp / ~15 socket | #209 to #215, all merged via #266, #267, #269 |
| 2 | One ACK per receive pass, drain loop, reassembly walk bound | socket 15 to 30, gen_udp 30 to 38 | #274 |
| 3 | GSO uniform-run splitting | 32 to 48, retransmits to zero, sendmsg 82k to 6k per 100 MiB | #217, merged via #267 |
| 4 | AEAD context NIF (EVP context per key), key-update counter fold | 48 to 59-68 | #216 (closed), carried as the first five commits of #285 |
| 5 | Chunk-run sender, delivery coalescing (opt-in), receive queue bound 512 | 59-66 to 76-81, zero retransmits | #282 (coalescing), #283 (run sender, queue bound) |
| 6 | Pacing unfreeze on sub-ms RTT, burst continuation off the timer wheel | 83 to 194 | #283 |
| 7 | Fused per-run seal (protect_run) and per-packet open (open_packet) in the NIF | send crypto 9% to 2% of connection CPU | #285 |
| 8 | In-NIF frame parser after batched decrypt (open_run) | 190 to 222 (+17%) | #285 |
| 9 | Fused GRO-train open, ordered reassembly buffer, fused receive bookkeeping, lean stream-data path | receive profile items off the top list; part of the path from 222 to 339 | ordered buffer, fused bookkeeping and lean path in #274; train open in #285 |
| 10 | Stream-train vectorization in the batched receive path | 339 to 405 (+19%) | #285 |
| | Plateau | 397 to 412 MiB/s at 1400-byte datagrams; msquic 803, Erlang TLS 481 to 557 on the same rig | |

The profile at the plateau: open_run 23% (about 1 us per packet, the AES-GCM hardware floor), recvmsg 7%, iolist flatten 4.5%, the rest a tail of train traversals. Going further needs C-side packet bookkeeping, jumbo datagrams (498 MiB/s measured) or io_uring, which we decided not to pursue.

### Server CPU at connection scale (50-site CT, same host, eq vs msquic)

| Step | Change | Server CPU seconds (eq / msquic) | Upstream |
|---|---|---|---|
| 0 | Before | 71 / 24 (2.9x), of which syscall time 26 s vs 3 s | |
| 1 | Shared sender process for server connections (serializes sendmsg on the shared socket; contention cost 98.6 vs 4.7 us per send at 50 concurrent senders) | 69.5 / 24, syscall time 25 to 17.7 s; fan-out model 24.3 to 11.0 s | #283 (pure Erlang) |
| 2 | Listener receive sweep, PTO re-arm deferral | 67.8 / 24 | #283 |
| 3 | sendmmsg NIF: the shared sender drains up to 64 batches into one syscall | 60.7 / 24 (2.5x), syscall time 24 to 10.3 s, sender reductions 68M to 10M per 10 s | fork only |
| 4 | Pacing continuation fix (step 6 above) | 58.6 / 24.4 (2.4x) | #283 |
| 5 | Keep-alive PING spin against a quiet peer | 24.6 / 24.6 (1.00x); memory 274 vs 267 MB (was 464 vs 296) | #221, merged via #271 |

Half of the CT premium turned out to be the keep-alive spin (#221), the other half the send path; with both fixed eq runs the 50-site test at msquic's CPU and memory.

### What this means for the C question

The AEAD context alone is worth about 10 MiB/s here (step 4), which matches your 9% figure. The larger NIF wins came from fusing work across a run of packets, not from moving AEAD alone: protect_run/open_packet (step 7), the in-NIF frame parser (step 8, +17%) and the GRO-train open with train vectorization (steps 9 and 10, +19% on top). Those need the context resource as a base, which is why they were queued behind #216. Without C the same path stops around 194 to 225 MiB/s; with it, about 400.

That fused path is now #285, ten commits on top of #283 with tests and CHANGELOG entries per commit, a pure-Erlang fallback that is today's code byte for byte, and a best-effort build. Measured the same way on the same machine with the stack as it is in the PRs (10 MB download sink, three interleaved rounds each):

| backend | #283, pure Erlang (MB/s) | #285, fused NIF path (MB/s) |
|---|---|---|
| gen_udp | 29.3 / 30.9 / 31.2 | 33.4 / 34.2 / 34.6 |
| socket | 39.8 / 40.6 / 42.7 | 58.5 / 60.4 / 61.4 |

gen_udp barely moves because it hands the connection one datagram at a time, so the run-based calls see runs of one; the GRO trains of the socket backend are what the fused path is for. The one piece still only in our fork is the sendmmsg NIF for the shared sender (another ~10% of server CPU at 50 connections); it can follow if #285 lands.

jbevemyr:
Updated the table above: every row now names its PR. The fused NIF path is #285 (on top of #274 → #282 → #283), measured against #283 with the same bench in the last table; #284 covers the interop runner.

PR #217, opened 2026-08-19
GSO-send mixed-size batches as uniform runs

jbevemyr:
The GSO flush gate is all-or-nothing: unless every packet in the batch (except the last) is exactly `gso_size`, the whole batch falls back to one sendmsg per packet. Bulk send batches almost always contain one odd-sized packet — an ACK or a flow-control update between full-sized data packets — so in practice most batches degrade to the per-packet path (profiled: ~82k sendmsg calls for ~82k packets despite 22-packet batches).

This splits a mixed batch into runs of equal-sized packets and sends each run in one `UDP_SEGMENT` sendmsg (with the run's own segment size), odd packets individually, preserving wire order exactly. A run may absorb one shorter trailing packet, since the kernel permits a short final segment. A partial kernel write falls back to segment-by-segment sends for that run's remainder (mirroring `flush_gso_partial`) and continues with the following runs.

Measured on a loopback bulk transfer (200 MiB single stream, socket backend): sendmsg count drops from ~82k to ~6k per 100 MiB and throughput improves ~50% at the point in our patch stack where this was measured, with zero retransmits.

### DIFF
--- a/src/quic_socket.erl
+++ b/src/quic_socket.erl
@@ -527,10 +527,13 @@ flush(#socket_state{gso_supported = true, batch_buffer = Buffer, gso_size = GSO}
     %% exactly gso_size. Handshake flights coalesce Initial-padded-to-
     %% 1200 with a ~400 byte Handshake packet; a naive GSO split
     %% mis-aligns those boundaries and the client cannot decode. When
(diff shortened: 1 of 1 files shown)


benoitc:

benoitc:

PR #218, opened 2026-08-19
Switch away from a retired DCID (RFC 9000 §5.1.2)

jbevemyr:
When a NEW_CONNECTION_ID frame's `retire_prior_to` covers the CID the client is actively sending with, `retire_peer_cids` sends RETIRE_CONNECTION_ID and prunes the pool entry — but `#state.dcid` keeps pointing at the retired CID, so every subsequent packet still uses it.

Against msquic this is fatal: msquic tolerates the retired CID briefly, then treats packets to it as unroutable and answers each with a **stateless reset** — whose token was stored on the pool entry that was just pruned, so `check_stateless_reset/2` can never match it. The connection goes irrecoverably deaf mid-flight.

Diagnosed in a live trace by logging the peer CID pool at decrypt-failure time: the client was sending with DCID `A385...` while the pool held only seq 2 and 3 (`retire_prior_to = 2` had pruned seq 0/1, including the entry for the in-use CID), and the arriving resets carried the token for the pruned CID.

Fix: before retiring, if the current DCID is among the to-be-retired entries, switch to an active CID from the pool (RFC 9000 §5.1.2: "An endpoint SHOULD supply a new connection ID when the peer retires a connection ID that the endpoint is using").

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -3955,13 +3955,35 @@ handle_packet_loop(Data, State) ->
             maybe_reenable_socket(State),
             State;
         {error, Reason} ->
-            %% Log decryption failure for debugging
+            %% Log decryption failure for debugging. Include the packet
+            %% source, our CID expectation and the datagram head/tail so
+            %% a misrouted datagram, a foreign connection's packet or a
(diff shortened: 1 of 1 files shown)


benoitc:

PR #219, opened 2026-08-20
Scale the pacing burst allowance with the pacing rate

jbevemyr:
pacing_max_burst is a fixed 12 packets while pacing wakeups (send_after) have ~1 ms timer resolution. Whenever the sender outruns the incoming ACK clock, the pacing timer becomes the only wakeup source, and the 12-packet bucket then caps throughput at 12 packets per wakeup (~17 KB/ms) regardless of the computed pacing rate. On a fast LAN this pins bulk transfers to a small fraction of the paced rate; we measured a hard plateau far below both the pacing rate and the congestion window, with the token bucket permanently empty.

This scales the allowance to 2 ms of the paced rate, floored at the old 12 * max_datagram_size. Micro-bursts stay bounded (the same idea as Linux fq's pacing quantum) while a 1 ms clock can sustain the configured rate. On RTTs above a few milliseconds the ACK clock dominates and behavior is unchanged.

On loopback this roughly doubles single-stream bulk throughput in our benchmarks. Full eunit passes.

Related to the performance series (#203-#218) but independent: applies directly to main.

### DIFF
--- a/src/quic_cc_newreno.erl
+++ b/src/quic_cc_newreno.erl
@@ -812,7 +812,7 @@ update_mtu(#cc_state{max_datagram_size = OldMDS, minimum_window = OldMinWin} = S
     NewMinimumWindow = max(2 * NewMTU, (OldMinWin * NewMTU) div OldMDS),
 
     %% Update pacing_max_burst (12 * max_datagram_size)
-    NewPacingMaxBurst = 12 * NewMTU,
+    NewPacingMaxBurst = max(12 * NewMTU, 2 * State#cc_state.pacing_rate),
 
     ?LOG_DEBUG(
         #{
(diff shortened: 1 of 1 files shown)


jbevemyr:
Correction: the "roughly doubles single-stream bulk throughput" claim in this description was measured with a benchmark we have since found to be flawed (it silently dropped data past the async send-queue cap; details in benoitc/erlang_quic#220). The analysis of the 12-packet burst cap itself stands (the token bucket does pin throughput to ~17 KB per pacing wakeup when the sender outruns the ACK clock), but the measured magnitude of the improvement needs re-measuring with our fixed, delivery-verified harness. I will follow up with honest numbers.

benoitc:

benoitc:

PR #223, opened 2026-08-24
Handle overlapping and duplicate chunks when reassembling

jbevemyr:
Both reassembly buffers look up the next bytes by exact offset. `process_crypto_buffer/2` does `maps:find(ExpectedOffset, Buffer)` and `extract_contiguous_data/3` does `maps:take(Offset, Buffer)`. That assumes a chunk always starts precisely where the last delivery ended, which the protocol does not promise: RFC 9000 §2.2 permits overlapping STREAM frames, and §13.3 lets a peer split or coalesce data it retransmits differently from the original send.

### The stall

When a chunk straddles the delivery point, its key stays below that point and is never looked up again:

```
buffered: {1000 => 500 bytes} {1200 => 500 bytes}
deliver from 1000 -> takes key 1000, advances to 1500
                     looks for key 1500, finds nothing
                     bytes 1500..1699 are held forever
```

The stream stalls permanently. On the CRYPTO side the handshake does, which is the same end state as the split-header bug in #222: one side considers itself connected while the other waits for data it is already holding.

### The leak

Chunks below the delivery point had the mirror problem: extraction only ever removed exact matches, so they were kept forever.

On a stream they were inserted into the buffer but counted as zero new bytes, so they grew without being charged against `?MAX_RECV_BUFFER_BYTES` — unbounded memory the cap could not see. On the CRYPTO side they *were* charged, so accumulated duplicate retransmissions could reach `?MAX_CRYPTO_BUFFER_BYTES` and close a healthy connection with `CRYPTO_BUFFER_EXCEEDED`.

### The fix

Both paths now share `trim_reassembly_buffer/2`, applied when the exact lookup misses: chunks ending at or before the delivery point are dropped, a chunk straddling it is trimmed and re-keyed to that point, and extraction retries once. The fast path, an exact hit, is untouched, so in-order delivery costs nothing extra; only a miss pays the fold, and a miss is what used to mean "give up".

Insertion goes through `keep_longest_chunk/3`, because plain `maps:put` let a shorter retransmission at an offset displace a longer chunk already buffered and silently drop the tail.

Receive-buffer accounting is now derived from the buffer itself rather than from new-minus-delivered. With overlap the old estimate drifted upward, since overlapping bytes counted as new but were never delivered twice, and that drift would eventually trip the cap on a stream holding nothing.

### Tests

`test/quic_reassembly_tests.erl` covers exact joins, gaps, straddling overlap, overlap spanning two chunks, fully contained chunks, pruning of delivered and duplicate chunks, trim idempotence, and a randomised pass that fragments a known 4000 byte stream arbitrarily and requires it to reassemble byte-exactly.

6 of the 13 fail on the current code. Full eunit suite clean with the fix (2276 tests, 0 failures).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -147,6 +147,9 @@
     check_send_queue_flow_control/4,
     test_check_flow_control/6,
     close_reason_to_code/1,
+    %% Out-of-order reassembly (RFC 9000 §2.2, §13.3)
+    extract_contiguous_data/2,
+    trim_reassembly_buffer/2,
     %% Migration frame classification (RFC 9000 Section 9.1)
     is_probing_frame/1,
     contains_non_probing_frame/1,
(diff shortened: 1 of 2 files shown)


benoitc:

PR #224, opened 2026-08-24
Give up on an unresponsive peer with a disconnect timeout

jbevemyr:
This is a feature proposal rather than a bug fix, so the default is the part worth arguing about. See the note at the end.

### The gap

When a peer disappears without closing, the only thing that ends the connection is the idle timeout. That is correct per RFC 9000, but it is a long time to hold a connection that is provably dead, and there is a case where it is worse than just slow: a peer that restarts and sends stateless resets we cannot recognise. We keep probing, it keeps resetting, and nothing resolves until the idle timeout expires. With `idle_timeout` set generously, or `infinity`, the connection never goes away at all.

msquic solves this with a 16 second disconnect timeout, and this takes the same approach: if ack-eliciting data has been in flight with no ACK at all for the timeout, declare the peer dead and close instead of probing on.

### Two commits

**1. The timeout itself.** `quic_loss` gains `last_progress/1`, the time of the last ACK that made forward progress. `quic_connection` closes with `disconnect_timeout` when data has been in flight past that point for longer than the configured window. Configurable via the `disconnect_timeout` option, `infinity` to disable.

**2. Checking it on its own timer.** The first version evaluated the condition only when a `pto_timeout` happened to fire, which turns out to be a poor sampling point: PTO backoff doubles on every retransmission, so by the time a peer has gone quiet the probe interval is seconds wide and the deadline is discovered well past it. Measured against 50 hard-killed peers, a nominal 16 s timeout fired after 24.4 s, where msquic noticed the same peers in 15.7 s. The timeout was not wrong, its sampling was.

The check now runs on a dedicated timer armed at `last_progress + disconnect_timeout`, re-armed lazily when a fire finds the deadline has moved. The delay is never zero, so it cannot spin the way a re-arm derived from a stale timestamp can: with data in flight and the deadline not reached the remainder is positive, and with nothing in flight a full interval is armed. Cost is one timer per connection firing at most once per interval while idle, against the keep-alive timer's once per second. The PTO path keeps its check, where it is now just an early exit.

On a control tower with 50 connections this removed 19.6 s from the phase where the peer process is killed and the server has to notice.

### About the default

The branch defaults to 16000 ms, which is what we ship and what msquic uses. That does change behaviour for existing users: a path that blackholes for 20 seconds survives today and would not after this. If you would rather keep it purely additive, say so and I will default it to `infinity` so it only applies when asked for. I do not have a strong view on which is right for upstream, only that the choice should be explicit rather than a side effect.

Full eunit suite clean on this branch (2263 tests, 0 failures).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -497,6 +497,10 @@
     %% peer then lingers for the whole idle timeout; msquic uses a 16 s
     %% disconnect timeout for the same reason).
     disconnect_timeout = 16000 :: pos_integer() | infinity,
+    %% Dedicated timer for the above. Checking it only when a PTO happened
+    %% to fire made the effective timeout the next PTO after the deadline,
+    %% and PTO backoff doubles, so a nominal 16 s took 24 s or more.
+    disconnect_timer :: reference() | undefined,
 
     %% Pacing (RFC 9002 Section 7.7)
     pacing_timer :: reference() | undefined,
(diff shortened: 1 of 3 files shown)


benoitc:

PR #225, opened 2026-08-24
Do not drop async sends on flow control or before connected

jbevemyr:
`send_data_async/4` is fire-and-forget, so any path that returns an error for it loses the payload, and the loss is invisible at both ends: the bytes never reach the wire, so the peer cannot detect a gap, and the next send on the stream simply continues from a later offset. The application stream ends up missing a chunk in the middle with nothing reported anywhere.

We hit this as permanent mid-stream message loss against an msquic peer under load, and it took a while to find precisely because nothing logs it.

Two paths did it.

### Flow control

`do_send_data/6` returned `{error, {flow_control_blocked, _}}` when the peer's window had no room:

```erlang
BlockedFrame = {data_blocked, MaxDataRemote},
_FinalState = send_frame(BlockedFrame, State),
{error, {flow_control_blocked, connection}};
```

It now queues instead, which is the same contract the congestion-control remainder path already uses: the offset is advanced at queue time and `process_send_queue/1` drains the entry when MAX_DATA or MAX_STREAM_DATA arrives.

Advancing the offset at queue time is the part worth reviewing. It is what makes later sends on the stream order behind the queued entry rather than overwrite its range. QUIC reassembly is offset-based, so the order the queued entries actually go out in does not matter.

The old comment said "RFC 9000: Don't queue data beyond flow control limits. Caller should retry after receiving MAX_STREAM_DATA." That is right for a caller that can retry. `send_data_async` cannot, and the synchronous path already queues on congestion control, so this makes the two consistent.

### Sends before `connected`

An async send arriving before the state machine reaches `connected` fell through to the `handle_common_event/4` catch-all and was discarded. The owner is told the connection is up a flight before the handshake concludes, so this is an ordinary race rather than misuse.

It now queues into `pending_data` the way the synchronous handshaking-state clause already does, flushed by `send_pending_data/2` on entering `connected`, bounded by `MAX_PENDING_DATA_ENTRIES` with a warning if that limit is reached. A bounded queue with a logged drop is a large improvement on an unbounded silent one.

### Tests

`test/quic_async_send_loss_tests.erl` covers `queue_blocked_send/6`'s contract: that a blocked send is queued rather than erroring, that the offset advances by the payload size, that FIN and empty payloads survive, and that iodata is measured correctly.

Being straight about the limits: these exercise the new function directly rather than driving `do_send_data/6`, which needs a full connection to reach that branch. The wiring itself is the two changed call sites in the diff.

Full eunit suite clean on this branch (2268 tests, 0 failures).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -146,6 +146,7 @@
     convert_rest_ranges/2,
     check_send_queue_flow_control/4,
     test_check_flow_control/6,
+    test_queue_blocked_send/5,
     close_reason_to_code/1,
     %% Migration frame classification (RFC 9000 Section 9.1)
     is_probing_frame/1,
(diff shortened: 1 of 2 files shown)


benoitc:

| | two dropped |
|---|---|
| main | fails |
| #252 alone | fails |
| #230 alone | fails |
| #252 + #230 | passes |

benoitc:

PR #228, opened 2026-08-25
interop: give the server the connection pid, not the DCID

jbevemyr:
The interop server never served a request, so the server side cannot have been passing any interop test case that requires a response.

Its connection handler is registered with arity 2, which `quic_listener` calls as `Fun(ConnPid, DCID)`. The second argument was named `Conn` and then used as the connection handle throughout: matched against in every `receive`, and passed to `quic:send_data/4` and `quic_connection:key_update/1`.

The messages carry the connection **pid**, so none of the receive patterns could match a DCID binary. Everything fell through to the catch-all, which logs and loops. A client connects, sends its request, and sits there until the idle timeout:

```
Handler got unexpected: {quic,<0.113.0>,{stream_opened,0}}
Handler got unexpected: {quic,<0.113.0>,{stream_data,0,<<"GET /test.txt\r\n">>,true}}
Handler got unexpected: {quic,<0.113.0>,{closed,idle_timeout}}
```

on the server, and `Some downloads failed` on the client.

This threads `ConnPid` through instead and drops the DCID, which the handler has no use for. Both interop images built from the same tree, on a docker network:

```
Downloading: https://iopserver:443/test.txt
Connected
Saved: /downloads/test.txt (8192 bytes)
All downloads successful
```

and the downloaded file is byte-identical to the source.

One thing worth deciding separately: `interop/README.md` lists all ten core test cases as passing for both client and server. That table cannot be accurate for the server column as things stand, and I have not touched it, since you may want to re-run the matrix yourself rather than take my word for which of the ten now pass. I have only verified `handshake` end to end.

Found while looking for a way to validate a stack of transport fixes against other QUIC implementations, which is what the interop runner is for.

### DIFF
--- a/src/interop/quic_interop_client.erl
+++ b/src/interop/quic_interop_client.erl
@@ -88,10 +88,11 @@ run_test(TestCase, RequestsStr, DownloadsDir) ->
             io:format("No requests specified~n"),
             halt(?EXIT_FAILURE);
         _ ->
-            Results = lists:map(
-                fun(Url) -> download_file(TestCase, Url, DownloadsDir) end,
-                Requests
-            ),
+            %% The runner asks for several files and requires them on ONE
+            %% connection: the transfer test counts handshakes and fails on
+            %% more than one, and multiplexing expects concurrent streams.
+            %% Connect once per host and reuse it for every path.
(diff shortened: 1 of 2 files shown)


benoitc:
Closing as contained in #245.

Review continues on #245.

PR #229, opened 2026-08-25
Send Version Negotiation for versions we do not support

jbevemyr:
quic_packet:encode_version_negotiation/3 was implemented but never
called: handle_initial_packet/5 took the version and never checked it,
so an Initial carrying an unknown version was routed into connection
setup and the peer got nothing back.

RFC 9000 Section 6.1 requires a server that receives a long-header packet
with an unsupported version to reply with a Version Negotiation packet
listing the versions it does support. Beyond the spec, probing with a
reserved version is how readiness checks decide a QUIC server is alive:
the QUIC Interop Runner's simulator does exactly this before it will
start a test, so the server never became reachable and every test case
was recorded as unsupported.

With this, the runner's probe answers in 2ms instead of timing out after
10s, and the handshake test passes against quic-go and ngtcp2.

Version 0 gets no reply, since that is itself a Version Negotiation
packet and answering one invites a loop.

Found while wiring erlang_quic up to the QUIC Interop Runner. Full eunit clean (2263 tests).

### DIFF
--- a/src/quic_listener.erl
+++ b/src/quic_listener.erl
@@ -691,7 +691,7 @@ get_tables(_) ->
 
 handle_packet(Packet, RemoteAddr, #listener_state{dcid_len = DCIDLen} = State) ->
     case parse_packet_header(Packet, DCIDLen) of
-        {initial, DCID, _SCID, Version, _Rest} ->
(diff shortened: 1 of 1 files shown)


benoitc:
Aggregated into #272; your commit is carried there.

PR #230, opened 2026-08-25
Send what fits when a write exceeds the peer's flow-control window

jbevemyr:
do_send_data/4 compared the whole write against the available window:

    case {DataSize =< ConnectionAllowed, DataSize =< StreamAllowed} of

so a write larger than the window took the blocked path and queued in
its entirety, putting nothing on the wire. That deadlocks the transfer.
The peer only extends the window as it consumes data, so sending nothing
means nothing ever arrives to open it, and both sides wait forever.

Measured against a peer advertising initial_max_stream_data of 524288:
transfers of 64K, 256K and 512K completed, 1M and 2M delivered zero
bytes and timed out. The sender logged send_data result: ok for the full
payload each time, because queueing had succeeded.

Now a write that only partly fits sends the part that does and queues
the remainder at the advanced offset, where process_send_queue/1 drains
it when MAX_DATA or MAX_STREAM_DATA arrives. Both 1M and 2M then
transfer byte-identically. The fully blocked case (no room at all) is
unchanged: BLOCKED frame plus queue.

Depends on queue_blocked_send/6 from the async-send fix, so this is
stacked on that branch.

**Stacked on #225** (needs `queue_blocked_send/6`); review that one first.

Found by running erlang_quic through the QUIC Interop Runner. Full eunit clean (2268 tests).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -146,6 +146,7 @@
     convert_rest_ranges/2,
     check_send_queue_flow_control/4,
     test_check_flow_control/6,
+    test_queue_blocked_send/5,
     close_reason_to_code/1,
     %% Migration frame classification (RFC 9000 Section 9.1)
     is_probing_frame/1,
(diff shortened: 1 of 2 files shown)


benoitc:

| | two dropped |
|---|---|
| main | fails |
| #252 alone | fails |
| #230 alone | fails |
| #252 + #230 | passes |

benoitc:

PR #231, opened 2026-08-25
Write TLS secrets to SSLKEYLOGFILE when it is set

jbevemyr:
Wireshark, and through it the QUIC Interop Runner, decrypts a capture by
reading traffic secrets from the file named by SSLKEYLOGFILE. Several
interop test cases refuse to grade a run without one: multiplexing, for
instance, returns "unsupported" before it looks at whether anything
transferred. erlang_quic mentioned the variable in the interop harness
comments but never wrote the file, so those cases could not pass however
well the transfer went.

quic_keylog writes the four TLS 1.3 lines in NSS Key Log format, keyed
by the ClientHello random. The server previously discarded that random
while parsing the ClientHello; it is now kept in tls_ch1_random, which
the client side already did.

This hands anyone who can read the file the ability to decrypt the
connection, so it stays off unless SSLKEYLOGFILE is set, and a line with
no client random is dropped rather than written unmatchable.

With this the runner grades multiplexing and transfer as passing where
both were previously unsupported or failed.

Found while running erlang_quic through the QUIC Interop Runner. Full eunit clean (2263 tests).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -3048,6 +3048,7 @@ send_server_handshake_flight(Cipher, _TranscriptHashAfterSH, State) ->
     ServerAppSecret = quic_crypto:derive_server_app_secret(
         Cipher, MasterSecret, TranscriptHashFinal
     ),
+    keylog_application(State#state.tls_ch1_random, ClientAppSecret, ServerAppSecret),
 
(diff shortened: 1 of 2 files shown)


benoitc:

PR #232, opened 2026-08-25
Send Retry when asked to, and let the server's cipher order be configured

jbevemyr:
Two options that existed but reached nothing.

The interop server set `retry => true` for the retry test case, while
quic_listener reads `address_validation`. Nothing consumed `retry`, so no
Retry packet was ever sent and the interop runner reported "Didn't find
any Retry packets". With the option corrected the retry test passes
against quic-go, where it previously could not run at all.

select_cipher/1 hardcoded the server's preference as [aes_128_gcm,
aes_256_gcm, chacha20_poly1305] and ignored any configuration, so the
`ciphers` option the interop server passes for the chacha20 case had
nowhere to take effect. The preference now comes from the connection
state, defaulting to the same order as before, so a deployment that must
not negotiate a given suite can say so.

Note this makes the chacha20 test reach the transport rather than fixing
it: with ChaCha20 negotiated, eq talks to eq fine but a quic-go client
stalls the connection. That is a separate interop problem and still open.

Found running erlang_quic through the QUIC Interop Runner. Full eunit clean (2263 tests).

### DIFF
--- a/src/interop/quic_interop_server.erl
+++ b/src/interop/quic_interop_server.erl
@@ -113,7 +113,9 @@ build_server_opts(TestCase, Cert, Key, WwwDir) ->
     %% Test case specific options
     case TestCase of
         "retry" ->
-            BaseOpts#{retry => true};
+            %% The listener reads address_validation; `retry' was set here
+            %% but nothing ever looked at it, so no Retry was sent.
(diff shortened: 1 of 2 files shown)


benoitc:

PR #233, opened 2026-08-25
Send the sendable prefix of a flow-control-blocked queued entry

jbevemyr:
The send queue drain treats flow control as all-or-nothing: if the head entry does not fit entirely within the connection and stream windows, it stays queued until a grant arrives that covers the whole entry.

A peer that extends its window incrementally as the application reads never issues such a grant. It sizes each `MAX_STREAM_DATA` to what it just consumed, so against a multi-megabyte queued entry the window grows in steps far smaller than the entry and the transfer wedges permanently.

Found with the QUIC Interop Runner. A quic-go client downloading from our server stalled at exactly the last offset it had granted, while ngtcp2 passed the same case because it opens a large window up front. quic-go's qlog showed it was sending window extensions the whole time; we were declining to use them.

The fix splits the head entry, sending the prefix the windows admit and putting the remainder back. Both parts go in via `pqueue_in_front` so the stream stays in offset order. Appending the remainder at the back would place it behind higher-offset entries of the same stream and strand it, which is the hazard that helper already documents.

This is the drain-path counterpart of #230, which fixes the same all-or-nothing defect on the direct write path.

Result in the interop runner, quic-go client against our server:

| | before | after |
|---|---|---|
| transfer | fail, stalls at the granted offset | pass |
| multiplexing | fail | pass |
| keyupdate | fail | pass |

`erlang-quic` against itself still passes transfer, multiplexing and handshake, and eunit is clean (2263 tests).

Based on #205, which introduces `pqueue_in_front` and `process_send_queue_unbudgeted`.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -8476,7 +8476,15 @@ queue_stream_data(
     } = State,
     Where
 ) ->
-    DataSize = iolist_size(Data),
+    %% Normalise to a binary: consumers like
+    %% dequeue_small_stream_frame_tuple/1 hand the entry's data straight
+    %% to quic_frame:encode/1, which requires a binary.
+    DataBin =
+        case is_binary(Data) of
+            true -> Data;
+            false -> iolist_to_binary(Data)
+        end,
+    DataSize = byte_size(DataBin),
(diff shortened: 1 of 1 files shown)


benoitc:

PR #234, opened 2026-08-25
Use the negotiated cipher when unprotecting short headers

jbevemyr:
decrypt_app_packet removed header protection via `unprotect_short_header/4`, which guesses the cipher from the HP key length. AES-256-GCM and ChaCha20-Poly1305 both use 32-byte HP keys, so on a ChaCha20 connection the guess lands on AES-256 and the computed mask is unrelated to the real one. The recovered packet number and length are garbage, AEAD open fails, and every received 1-RTT packet is silently dropped.

The failure is invisible until interop: the long-header path passes the cipher explicitly, so the handshake completes, and the send path also knows the cipher, so the peer decrypts our packets fine. The connection then sits in a one-way conversation until it idles out. Found with the QUIC Interop Runner, where the chacha20 case failed against every other implementation.

Add `unprotect_short_header/5` taking the cipher and pass the one from the negotiated keys; `/4` stays as a delegating wrapper.

The RFC 9001 Appendix A.5 ChaCha20-Poly1305 short-header vector is added as a regression test; it fails against the old inference path (the old code recovers header `4972AF` where the RFC expects `4200BFF4`).

Verified with the interop runner: with this fix the chacha20 case passes with quic-go, ngtcp2 and picoquic clients against our server. (The client role needs one more fix — the ClientHello offers a hardcoded cipher list — which comes separately as it builds on #232.)

### DIFF
--- a/src/quic_aead.erl
+++ b/src/quic_aead.erl
@@ -46,6 +46,7 @@
     unprotect_long_packet/7,
     %% 2-stage short header receive API (for key-phase handling)
     unprotect_short_header/4,
+    unprotect_short_header/5,
     decrypt_short_payload/8
 ]).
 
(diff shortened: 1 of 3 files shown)


benoitc:

PR #235, opened 2026-08-25
Offer the configured cipher suites in the ClientHello

jbevemyr:
The client's ClientHello offered a hardcoded `TLS_AES_128_GCM_SHA256`, regardless of the `ciphers` option. The rest of the client is already cipher-aware: the key schedule follows the ServerHello's selection, including the SHA-384 schedule for AES-256. Only the offer was frozen, so the server's choice was never really a choice, and a peer that requires another suite failed the handshake with no common cipher.

Found with the QUIC Interop Runner's chacha20 case, where the server side only accepts ChaCha20-Poly1305 and closed the handshake with a handshake_failure alert (0x128).

Build the suite list from the `ciphers` option, defaulting to all three RFC 9001 suites, and validate the ServerHello against it: RFC 8446 §4.1.3 requires the selected suite to be one the client offered, so a selection outside the list now aborts with `illegal_parameter` instead of silently proceeding with a suite we never advertised.

Together with #234 this makes the interop chacha20 case pass in both roles: verified against quic-go, ngtcp2 and quiche servers as client, and quic-go/ngtcp2 clients against our server, with handshake and transfer still green everywhere.

Based on #232, which introduces the `ciphers` option and `cipher_preference` state.

### DIFF
--- a/src/interop/quic_interop_server.erl
+++ b/src/interop/quic_interop_server.erl
@@ -113,7 +113,9 @@ build_server_opts(TestCase, Cert, Key, WwwDir) ->
     %% Test case specific options
     case TestCase of
         "retry" ->
-            BaseOpts#{retry => true};
+            %% The listener reads address_validation; `retry' was set here
+            %% but nothing ever looked at it, so no Retry was sent.
+            BaseOpts#{address_validation => always};
         "chacha20" ->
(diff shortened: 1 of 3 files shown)


benoitc:

PR #236, opened 2026-08-25
Reclaim streams when the FIN is acked, not when it is sent

jbevemyr:
The send side of a stream was marked done as soon as its FIN-bearing write went out, and for writes that had to be queued it was never marked at all. Both halves are wrong in opposite directions.

**Marking at send time** reclaims the stream, and returns MAX_STREAMS credit, before the peer has received anything. Against a client that opens streams as fast as credit arrives, the server hands out the whole request set almost immediately: in the interop multiplexing case quiche accumulated over a thousand outstanding requests, hit its per-process file descriptor limit, and fell over. RFC 9000 §3.1 ends the send side at "Data Recvd", which requires acknowledgment; granting on ack paces credit to what the peer actually completes, which is also what quic-go and ngtcp2 do.

**Never marking a queued FIN** leaked the stream: a response that hit the congestion or pacing gate was drained later without any bookkeeping, so the stream stayed in the map forever and its MAX_STREAMS credit was never returned. All hundred initial streams could wedge this way at once (easy to hit when many requests arrive in one burst), ending the connection at the initial stream limit.

The ack path now closes the send side when a FIN-bearing STREAM frame is acked and nothing below the final offset remains queued or in flight, reusing the reset-at machinery's checks; a stream whose earlier bytes are still outstanding is parked and finished by a sweep on a later ack. The drain path records `send_fin` when a queued FIN finally goes out, with a queue-count check so a re-queued zero-length FIN entry is not mistaken for a sent one.

Verified with the interop runner: multiplexing passes with quic-go, ngtcp2 and picoquic clients against our server and for our client against quic-go, with handshake/transfer green throughout. eunit clean (2263 tests).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -418,6 +418,8 @@
     %% StreamId => send reliable size, for local RESET_STREAM_AT streams whose
     %% data below the reliable size is not yet fully acked. Drained as acks arrive.
     pending_send_reset_at = #{} :: #{non_neg_integer() => non_neg_integer()},
+    %% FIN acked but earlier bytes still queued or in flight
+    pending_fin_reclaim = #{} :: #{non_neg_integer() => non_neg_integer()},
     %% Lost control-frame retransmissions deferred by congestion control, replayed
     %% through the CC-checked retransmit path when cwnd reopens.
     deferred_ctrl_retransmits = [] :: [term()],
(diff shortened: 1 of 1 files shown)


benoitc:
Tried to build a local repro for both halves of this and could not fully, so recording what I have rather than leaving it as a silent gap.

**What I did reproduce, and it supports the change.** Opening streams as fast as credit arrives and writing 24 KB with FIN on each, without reading responses, against the in-tree echo server:

- main: runs ahead to 203 streams before `stream_limit`
- this branch: 197

That is the "Data Recvd" change doing exactly what it says. Credit granted on acknowledgment instead of at send time paces the peer to what it has actually completed, and the run-ahead shrinks. Small numbers here because the echo server completes quickly; against a peer that opens as fast as credit arrives and has an fd limit, that gap is the difference you saw with quiche.

Worth flagging for anyone else writing a test here: my first attempt asserted that all 250 streams could be opened in that burst, which is backwards. Demanding unbounded run-ahead penalises the fix. A useful case has to complete its streams, not race ahead of them.

**What I could not reproduce: the queued-FIN leak.** I drove 250 streams in completing batches of 50, each batch written before any of it was read so the writes sit behind the congestion and pacing gates, then drained fully before the next. Both main and this branch complete 250 of 250. Sequential one-at-a-time also completes on both.

So the leak needs conditions the echo server does not create, presumably a response large or slow enough that the FIN genuinely queues and stays queued. If you have a shape in mind that forces it I will write the case; otherwise the interop multiplexing run is the evidence, and it is decent evidence.

Not superseding this PR, since I have nothing to add to it beyond the above.

jbevemyr:
The shape that forces the queued-FIN leak is the server's response FIN queuing at write time; the observable is MAX_STREAMS starvation for the peer opening the streams.

Two things about the batch repro made it invisible. The writes sitting behind the gates were the client's, and a leaked client FIN only shows up as the client's own stream map growing; the stream-limit credit the client waits on is returned by the server, keyed to the server's response send side. And on loopback the echo responses mostly go out inline once the cwnd opens, so the response FIN rarely lands in the queue at all.

To force it deterministically, make the client grant a stream window smaller than the response: connect with an initial bidi max_stream_data of 16 KB against a server that responds with 64 KB. Every response then blocks on stream flow control mid-write, so the FIN is queued at write time with certainty. The window auto-extends as data is delivered, the response drains, and on main the drained FIN is never recorded: the transfer succeeds from the application's point of view, the stream never closes server-side, and its MAX_STREAMS credit is never returned. Open-and-complete streams one at a time this way and main wedges exactly at the initial bidi stream limit with every transfer having succeeded; this branch keeps going. Nothing in the case is timing-based: the queuing is guaranteed by the window, not by racing the cwnd.

A cwnd-based version (responses much larger than the initial window, many at once) also reproduces it but only statistically; the flow-control version pins the drain path every time.

benoitc:

PR #237, opened 2026-08-25
interop: buffer fragmented requests, poll for stream credit

jbevemyr:
Two interop-harness fixes found running the multiplexing case, one per role.

**Server** (first commit): the handler treated every `stream_data` delivery as a complete request. A request split across STREAM frames was served once per fragment: the first piece parsed as a plausible path (`parse_request` accepts a line without CRLF) and got a 404 with FIN, the rest got a 400 with a second FIN at a higher offset. Redeclaring the final size on one stream is a protocol violation, and picoquic rightly closed the whole connection with `FINAL_SIZE_ERROR` (0x6). Buffer per stream and parse on `fin=true`.

**Client** (second commit): `collect_loop` refilled from the pending list only on stream completion events. MAX_STREAMS arriving at the connection process produces no message to the owner, so when the last grant lands after the last completion has been handled the client sits in receive forever; against quic-go it stopped at exactly the initial 100 streams. Poll with a short timeout while requests wait for credit. Requests that never got a stream were also invisible in the result, so a run that fetched 100 files of 1999 still printed "All downloads successful"; they now count as failures.

With these plus #236, multiplexing passes with quic-go, ngtcp2 and picoquic clients against our server, and for our client against quic-go. The remaining multiplexing failures in the matrix are external: the quiche client keeps every dump file open until the connection ends and panics on EMFILE at ~1019 requests (also against quic-go), and the ngtcp2 server image aborts on a UBSan null-pointer check in std::format mid-run (also with a picoquic client).

Based on #228.

### DIFF
--- a/src/interop/quic_interop_client.erl
+++ b/src/interop/quic_interop_client.erl
@@ -88,10 +88,11 @@ run_test(TestCase, RequestsStr, DownloadsDir) ->
             io:format("No requests specified~n"),
             halt(?EXIT_FAILURE);
         _ ->
-            Results = lists:map(
-                fun(Url) -> download_file(TestCase, Url, DownloadsDir) end,
-                Requests
-            ),
+            %% The runner asks for several files and requires them on ONE
+            %% connection: the transfer test counts handshakes and fails on
+            %% more than one, and multiplexing expects concurrent streams.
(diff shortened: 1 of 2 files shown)


benoitc:
Closing as contained in #245.

Review continues on #245.

PR #238, opened 2026-08-25
Select the PSK from a session ticket instead of only feeding 0-RTT

jbevemyr:
A ClientHello offering a resumption ticket validated the ticket and the binder, but used the result only to derive 0-RTT receive keys. The handshake itself continued on a zero PSK down the full certificate path, so resumption never actually resumed: the interop runner failed the case with "Server sent a Certificate message in the second handshake", and a client offering a ticket without early data got nothing at all out of it. The code's own comment described the gap.

A validated ticket now selects the PSK for the handshake (psk_dhe_ke), which flows through the existing external-PSK machinery: the ServerHello echoes `selected_psk_identity` and the certificate flight is skipped. The early-keys derivation stays, gated on the client actually offering early data. An unknown or expired ticket still falls back to a full handshake per RFC 8446 §4.2.11.

Verified with the interop runner together with #238/#239/#240/#241: resumption and zerortt pass for our server with quic-go, ngtcp2 and quiche clients, and for our client against quic-go, ngtcp2 and picoquic servers. eunit clean (2263 tests).

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -621,6 +621,7 @@
         undefined
         | #{
             identity => binary(),
+            identity_idx => non_neg_integer(),
             secret => binary(),
             mode => psk_dhe_ke | psk_ke
         },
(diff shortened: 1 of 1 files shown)


benoitc:

PR #240, opened 2026-08-25
0-RTT: only the client sends it, and lost 0-RTT data is resent

jbevemyr:
Two 0-RTT defects, both found with the interop runner's zerortt case.

**The handshaking-state send path chose 0-RTT whenever early keys were present, for either role.** A server also holds early keys, but only for receiving: a handler answering early data before the handshake completed had its response encrypted as a 0-RTT packet, which RFC 9000 §17.2.3 reserves for the client and which the client can never decrypt. The response vanished and the client eventually re-requested in 1-RTT. The server now queues such writes like any other pre-handshake data; they go out as 1-RTT when the connection completes.

**0-RTT packets were sent without being registered with loss detection**: `next_pn` was bumped and the datagram written, nothing else. A dropped 0-RTT request was never retransmitted, so its stream hung forever. Blasting forty requests into the simulator's 25-packet queue reliably lost several this way, every run. 0-RTT shares the application number space, so the packets are now registered there like any 1-RTT send; loss recovery then re-frames them into 1-RTT automatically. An app-space PTO can now fire before 1-RTT keys exist, so the probe path sits out until the keys arrive; handshake-space recovery drives progress meanwhile.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -1799,26 +1799,33 @@ handshaking({call, From}, open_stream, #state{early_keys = _EarlyKeys} = State) 
 handshaking(
     {call, From},
     {send_data, StreamId, Data, Fin},
-    #state{
-        early_keys = undefined,
-        pending_data = Pending
-    } = State
+    #state{role = client, early_keys = EarlyKeys} = State
+) when EarlyKeys =/= undefined ->
+    %% Send as 0-RTT data. 0-RTT sending is a client-only affair (RFC
+    %% 9000 §17.2.3): a server also holds early keys, but only for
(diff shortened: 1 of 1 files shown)


jbevemyr:
Pushed a second commit: the idle state had the same ungated 0-RTT send clause as handshaking (a server answering early data from idle still sent 0-RTT packets), and the 0-RTT send path had no fragmentation — a 10 KB write left as a single IP-fragmented datagram (RFC 9000 §14). Found via the interop resumption case with a picoquic client, whose request rides in early data; with this, resumption and zerortt pass with picoquic against our server.

benoitc:

PR #241, opened 2026-08-25
Stop sending MAX_STREAM_DATA for streams whose final size is known

jbevemyr:
The receive-side window update triggered purely on headroom shrinking below half the maximum window. When the configured initial stream window is much smaller than `fc_max_receive_window`, that condition is true from the first delivered byte, so every completed stream got a parting MAX_STREAM_DATA it can never use: RFC 9000 §19.10 notes the frame is pointless once the final size is known.

In request/response traffic this is one wasted frame per stream. The interop zerortt case made it visible: the grader budgets the client's total 1-RTT bytes, and forty of these frames were a fifth of it.


benoitc:

PR #242, opened 2026-08-25
interop: make the resumption and zerortt cases test what they claim

jbevemyr:
Harness fixes for the two ticket-based cases, found while bringing the transport-side resumption work (#238, #239, #240) through the runner.

**Client:**
- The ticket file lived under `/downloads`, which the runner diffs against the requested files; the stray file failed an otherwise passing run.
- Both cases fetched the same file in both phases. The runner hands two (resumption) or forty (zerortt) files and expects the first on the fresh connection and the rest on the resumed one.
- Phase 1 wrote only the final `stream_data` chunk to disk, truncating any file larger than one frame.
- The zerortt case now fetches the whole request set on the 0-RTT connection through `collect_streams`, so requests go out as early data and credit-blocked opens are retried; the first connection only fetches a ticket. PMTU probing is disabled on both connections: the grader budgets the client's total 1-RTT bytes, and padded probes alone blew it.

**Server:**
- Start the quic application: without its supervision the resumption ticket table is owned by the first connection that created it and dies with it (#239).

With the transport PRs in place: resumption and zerortt pass for eq against itself, our server with quic-go/ngtcp2/quiche clients, and our client against quic-go/ngtcp2/picoquic servers. Based on #237.

### DIFF
--- a/src/interop/quic_interop_client.erl
+++ b/src/interop/quic_interop_client.erl
@@ -88,10 +88,11 @@ run_test(TestCase, RequestsStr, DownloadsDir) ->
             io:format("No requests specified~n"),
             halt(?EXIT_FAILURE);
         _ ->
-            Results = lists:map(
-                fun(Url) -> download_file(TestCase, Url, DownloadsDir) end,
-                Requests
-            ),
+            %% The runner asks for several files and requires them on ONE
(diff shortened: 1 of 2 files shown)


benoitc:
Closing as contained in #245.

Review continues on #245.

PR #243, opened 2026-08-25
QUIC v2: correct wire format and RFC 9368 compatible negotiation

jbevemyr:
The v2 support was self-consistent but not RFC 9369: it interoperated with itself and nothing else. Three classes of fix, found with the QUIC Interop Runner's v2 case.

**Wire format**: the long-header packet type bits are version-specific (RFC 9369 §3.2 rotates them, deliberately), but `type_to_bits` and every hand-built first byte used the v1 values for all versions. The v2 salt and the Retry integrity key/nonce were pre-RFC draft values, and the HKDF labels for packet protection and key update stayed `"quic *"` where v2 requires `"quicv2 *"` (§3.3). All of these are now selected by version, with the negotiated version threaded through key derivation. (The draft-era constants were a perfect trap: eq↔eq v2 worked, every other implementation failed.)

**Negotiation (RFC 9368)**: the `version_information` transport parameter (0x11) is encoded and decoded, the client advertises the versions it accepts (a new `versions` option, preference order), and the server picks its most preferred version from the client's set before any version-dependent derivation, skipping the switch when the client offers 0-RTT (early data is bound to the first flight's version). Both sides re-derive Initial keys for the negotiated version and keep the pre-switch set so first-flight packets still decrypt. The client adopts the server's version from the first long-header packet, gated on having offered it.

The interop client's v2 case now starts in v1 offering `[v2, v1]` — the order matters against servers like picoquic that honour the client's preference.

Interop runner: v2 passes for erlang-quic against itself, against ngtcp2 in both roles, and against picoquic as server. (picoquic as client against our server is still under investigation.) Handshake/transfer regressions: none; eunit clean (2263 tests).

### DIFF
--- a/include/quic.hrl
+++ b/include/quic.hrl
@@ -142,6 +142,8 @@
 -define(TP_ACTIVE_CONNECTION_ID_LIMIT, 16#0e).
 -define(TP_INITIAL_SCID, 16#0f).
 -define(TP_RETRY_SCID, 16#10).
+%% RFC 9368 version_information
+-define(TP_VERSION_INFORMATION, 16#11).
 
 %% RFC 9221 - QUIC Datagrams
 -define(TP_MAX_DATAGRAM_FRAME_SIZE, 16#20).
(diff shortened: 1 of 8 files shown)


jbevemyr:
Pushed a second commit: the negotiation ran after the HelloRetryRequest decision, so a client needing an HRR (picoquic offers a P-256 key share first) got the HRR in v1 and the ServerHello in v2 — a mid-handshake version switch it rightly rejected. Negotiation now settles on the first ClientHello, before cipher/group selection. With this and the group-preference fix, v2 passes with picoquic in both roles too, completing the matrix: v2 green against ngtcp2 and picoquic in both roles plus eq↔eq.

benoitc:
Aggregated into #272; your commit is carried there.

PR #244, opened 2026-08-26
h3: end the message body at stream end, not at a frame boundary

jbevemyr:
A request-stream FIN that arrived in a bare final QUIC frame — after the last complete H3 frame rather than fin-marking its final chunk — was silently absorbed: the frame parser saw an empty buffer, returned "need more", and no fin-marked delivery ever reached the application. The body sat complete in the owner's hands with no way to know it had ended.

RFC 9114 §4.1 ends the message when the stream ends. quic-go's server sends its FIN exactly this way, so every eq-client H3 download from quic-go hung until timeout; implementations that fin-mark the last DATA chunk (including our own server) masked the gap — which made this a fun one to find: the interop http3 case passed against ngtcp2, quiche and picoquic and failed only against quic-go.

On a FIN with a cleanly drained buffer the stream now gets a final empty fin-marked delivery, unless a fin-marked frame already closed the body. A FIN arriving with a partial frame still buffered is now a connection error (`H3_FRAME_ERROR`, stream ended mid-frame) instead of silence, per RFC 9114 §7.1.

The regression test drives the client FSM with a fake QUIC connection: HEADERS, one complete DATA frame without fin, then a bare FIN; it fails against the old code.

With this (plus an interop-harness http3 case that comes separately), the interop runner's http3 test passes for erlang-quic in both roles against quic-go, ngtcp2, quiche and picoquic.

### DIFF
--- a/src/h3/quic_h3_connection.erl
+++ b/src/h3/quic_h3_connection.erl
@@ -2542,17 +2542,40 @@ handle_request_stream_data(
     Combined = <<Buffer/binary, Data/binary>>,
 
     case process_request_frames(StreamId, Combined, Fin, Stream, State) of
+        {ok, Rest, _Stream1, _State1} when Fin, Rest =/= <<>> ->
+            %% Stream ended inside an H3 frame: truncated (RFC 9114 §7.1).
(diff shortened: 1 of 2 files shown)


benoitc:
Aggregated into #273; your commit is carried there.

PR #245, opened 2026-08-26
interop: speak real HTTP/3 in the http3 case

jbevemyr:
The harness spoke hq-interop for every case, so the runner graded http3 as unsupported. This wires the existing `quic_h3` API up for it.

**Server**: the http3 case starts `quic_h3:start_server` with a handler serving the www directory; every other case keeps the bare listener and hq-interop.

**Client**: the http3 case fetches all files over one connection (the grader requires exactly one handshake) with GET requests through `quic_h3`, collecting response bodies by stream. The quic application is now started: the h3 path needs its supervision, while the hq path merely tolerated its absence.

With #244 (the bare-FIN body-end fix), the runner's http3 case passes for erlang-quic in both roles against quic-go, ngtcp2, quiche and picoquic, and against itself.

Based on #242.

### DIFF
--- a/src/interop/quic_interop_client.erl
+++ b/src/interop/quic_interop_client.erl
@@ -88,10 +88,11 @@ run_test(TestCase, RequestsStr, DownloadsDir) ->
             io:format("No requests specified~n"),
             halt(?EXIT_FAILURE);
         _ ->
-            Results = lists:map(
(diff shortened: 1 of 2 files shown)


benoitc:
Aggregated into #273; your commit is carried there.

PR #246, opened 2026-08-26
Accept all supported groups by default on the server

jbevemyr:
The server's default group preference held only x25519, so a client whose key_share leads with another curve was sent a HelloRetryRequest even though the crypto layer speaks secp256r1 and secp384r1 just fine. picoquic shares P-256 first, which put an HRR round trip — a full extra RTT — in front of every connection from it, and dragged the rarely-exercised HRR path into flows that did not need it (the HRR + compatible-version-negotiation interaction in #243 was found this way).

Default to every classical group the crypto layer supports. Clients that share x25519 are picked directly as before; an explicit `groups` option still restricts the set.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -1136,7 +1136,11 @@ init({server, Opts}) ->
             psk_callback => maps:get(psk_callback, Opts, undefined),
             psks => maps:get(psks, Opts, undefined)
         },
-        tls_groups = maps:get(groups, Opts, [x25519]),
+        %% Accept every group the crypto layer supports: a preference
+        %% list holding only x25519 forced a HelloRetryRequest round
+        %% trip on any client whose key_share leads with another group
+        %% (picoquic shares P-256 first).
+        tls_groups = maps:get(groups, Opts, [x25519, secp256r1, secp384r1]),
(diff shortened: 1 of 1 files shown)


benoitc:
Confirmed and carried into #257 with tests, authorship intact.

Measured the cost over the socket adapter rather than taking it on trust. Against a default server, counting the Initial-level datagrams the client sends before the connection is up:

| client `groups` | before | after |
|---|---|---|
| `[secp256r1, x25519]` | x25519, 5 | secp256r1, 3 |
| `[secp384r1, x25519]` | x25519, 5 | secp384r1, 3 |
| `[x25519]` | x25519, 3 | x25519, 3 |

So it is a full extra round trip, exactly as you described, and the server was also overriding the client's lead group rather than just adding a retry.

Two of the five new cases fail without the change. Three are fences that hold either way: x25519 clients keep their direct acceptance, an explicit `groups` option still restricts and still forces the HRR, and a disjoint client is still rejected. `quic_hrr_e2e_SUITE` passes unchanged, which is what checks that broadening the default did not quietly disable the HRR path.

Will close this once #257 lands.

benoitc:
Duplicate of #257, which carries this work with tests. Closing in favour of it.

PR #247, opened 2026-08-26
Handle SHA-384 suites in the Finished verification path

jbevemyr:
`parse_finished` truncated verify_data to 32 bytes (SHA-256 hard-coded) while `verify_finished/4` computes the full-length HMAC for the negotiated cipher. Any server that selects `TLS_AES_256_GCM_SHA384` (msquic does by OpenSSL default preference) therefore crashed the client with `badarg` from `crypto:hash_equals/2` on every Finished message, and the handshake could never complete.

This was masked while the ClientHello offered only `TLS_AES_128_GCM_SHA256`; once the offer includes `aes_256_gcm` (#235) a SHA-384 selection becomes reachable.

Fix:
- `parse_finished` returns the whole message body instead of truncating to 32 bytes.
- `verify_finished/3,4` and the PSK binder checks compare lengths before `crypto:hash_equals/2`, so a wrong-length verify_data or binder (the binder is attacker-controlled wire data) is a verification failure instead of a process crash.

Repro: connect an eq client with `ciphers => [aes_256_gcm]` to any eq or msquic server; before this change the connection never reaches `connected`. Regression tests included; full eunit, lint and xref pass.

### DIFF
--- a/src/quic_tls.erl
+++ b/src/quic_tls.erl
@@ -586,9 +586,10 @@ parse_certificate_verify(_) ->
 %% @doc Parse Finished message.
 -spec parse_finished(binary()) -> {ok, binary()} | {error, term()}.
 parse_finished(VerifyData) when byte_size(VerifyData) >= 32 ->
-    %% SHA-256 produces 32 bytes
-    <<Data:32/binary, _/binary>> = VerifyData,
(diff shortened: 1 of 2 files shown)


benoitc:

PR #248, opened 2026-08-26
Client socket backend: GRO receive and train-sized owner messages

jbevemyr:
The opt-in client socket backend never enabled `UDP_GRO`, so a GSO-batching peer's trains were segmented back into one `recvfrom` per datagram. That `recvfrom` also passed buffer size 0, whose 8 KiB default silently truncates any train that does arrive coalesced (`recv_gro` had the same latent truncation via `recvmsg(_, 0, ...)`).

This PR:
- enables GRO on the client socket when the platform supports it (per-socket `UDP_GRO`, detected the same way as the listener path),
- reads with `recv_gro` and a train-sized 64 KiB buffer,
- forwards a multi-packet train to the connection as a single `{udp_batch, ...}` message, processed in one receive pass — mirroring the listener's batched `{quic_packets, ...}` delivery.

The connection mailbox then holds trains instead of packets, and per-event flushes amortize over the train.

Measured on loopback bulk transfer (socket backend both ends, 4 MB x 64 streams): ~90 MB/s -> ~260 MB/s, with the win coming from ~46x fewer receive syscalls and messages.

Full eunit (2263), lint, xref and dialyzer pass.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -1732,6 +1732,10 @@ idle(info, {udp, Socket, _IP, _Port, Data}, #state{socket = Socket} = State) ->
     %% Flush batched packets and timers after processing incoming data
     FlushedState = flush_dirty_timers(flush_socket_batch(NewState)),
     check_state_transition(idle, FlushedState);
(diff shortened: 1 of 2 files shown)


benoitc:

PR #249, opened 2026-08-26
Retransmission congestion policy: exempt PTO probes, bind loss retransmissions to cwnd (RFC 9002 §7)

jbevemyr:
RFC 9002 §7.5: probe packets are not limited by the congestion window. The probe path went through `can_send_control` (cwnd + 1200-byte allowance), so once `bytes_in_flight` sat at or just above cwnd, every PTO probe was denied. No probe means no ACK, no ACK means loss detection never runs, and the connection deadlocks with a full sent_q until idle timeout.

Deterministic reproducer: a 10 MiB transfer through a UDP proxy that blacks out both directions for 2 s mid-transfer. The in-flight window is lost, `bytes_in_flight` fills cwnd exactly, and every subsequent PTO probe is refused: the transfer died on idle timeout in every run. With the exemption it completes in ~4.5 s. This also turned the interop runner's `blackhole` case from a hard fail into a pass.

Loss retransmissions (mode `normal`) stay subject to cwnd as before. Full eunit passes.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -166,6 +166,8 @@
     maybe_store_initial_reset_token/2,
     check_stateless_reset/2,
     test_state_for_reset/3,
+    %% Retransmission congestion policy (RFC 9002 §7)
+    retransmit_cc_allowed/3,
     %% NEW_TOKEN frame dispatch (RFC 9000 §8.1.3)
     process_frame/3,
     test_state_for_role/1,
@@ -9388,12 +9390,7 @@ send_retransmit_frames_cc(Frames, #state{cc_state = CCState, retransmits = R} = 
     Payload = iolist_to_binary([quic_frame:encode(F) || F <- Frames]),
     PacketSize = byte_size(Payload) + 50,
 
-    Allowed =
-        case Mode of
-            probe -> true;
-            normal -> quic_cc:can_send_control(CCState, PacketSize)
-        end,
-    case Allowed of
+    case retransmit_cc_allowed(Mode, CCState, PacketSize) of
         true ->
             send_app_packet_internal(Payload, Frames, State#state{retransmits = R + 1});
         false ->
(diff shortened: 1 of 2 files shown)


benoitc:
Tested this against the stall it describes.

I built a repro on the socket adapter: blackhole the client for 1.5s so a full window is stranded, then restore the path (`quic_stranded_window_recovery_SUITE` in #256). On main the transfer never recovers, 0 of 262144 bytes in 20s. This PR does not change that outcome, stable over 3 runs. #212, #206 and #223 are the same. The one that fixes it is #250, now carried in #256.

So the case for this change is RFC conformance rather than that stall, and it is still a real case.

RFC 9002 section 7: an endpoint MUST NOT exceed cwnd "unless the packet is sent on a PTO timer expiration or when entering a recovery period". That is two rules:

- a PTO probe is exempt outright
- an ordinary loss retransmission is not, so it stays bound by cwnd

main uses `can_send_control` for both, so it is wrong in both directions: it under-restricts loss retransmissions and over-restricts probes. This PR gets `probe` right and leaves `normal` on the lenient allowance. #212 does the opposite, `normal -> can_send` is right but `probe -> can_send_control` still gates the probe.

Neither is complete on its own. The union:

```erlang
Allowed =
    case Mode of
        probe -> true;
        normal -> quic_cc:can_send(CCState, PacketSize)
    end,
```

Both PRs make the same arity-2 to arity-3 refactor, so they conflict and only one can land. Could you fold the union into this one and close #212? Happy to take it the other way round if you prefer.

One thing worth pinning either way: I could not construct a case where the congestion gate alone produces visible harm, so whichever lands would benefit from a test at the policy level, asserting a probe is emitted above cwnd and a loss retransmission is refused above it.

jbevemyr:
Folded the union in here, exactly as you wrote it: `probe -> true` stays, `normal` now checks `quic_cc:can_send/2` instead of the lenient control allowance. Closing #212 in favour of this PR.

On the test: I extracted the decision into `retransmit_cc_allowed/3` and pinned it there in `quic_retransmit_cc_policy_tests`, since like you I could not construct end-to-end harm from the congestion gate alone. Four cases: a probe is allowed over a full window, a loss retransmission is refused over it, passes an open one, and the gate is cwnd proper rather than the control allowance (asserted whenever the two disagree on the same state). The refused path is already covered behaviourally: a denied retransmission lands in the deferred-retransmit queues and replays when cwnd reopens, which quic_flow_control_SUITE exercises.

Agreed on the #256 outcome as well: this one is conformance, #250 is what fixes the stranded-window stall.

benoitc:

PR #250, opened 2026-08-26
Do not apply Initial/Handshake ACKs to the 1-RTT loss tracker

jbevemyr:
Only 1-RTT packets are registered in the loss tracker, but ACK frames from every packet-number space were processed against it, and packet numbers restart per space. A Handshake-space ACK of PN 0..N therefore silently acked the first N 1-RTT packets out of sent_q; if those carried data the peer never received, nothing retransmitted them, and the peer stalled on a permanent stream hole until idle timeout.

The overlap window is exactly connection start, where a server's first 1-RTT response packets share numbers with the client's Handshake ACKs, so the typical casualty is a small request/response exchange. Measured with a 30%-loss UDP proxy and 1 KiB request/response per connection: ~5% of exchanges wedged (93-97 of 100); with the fix, 148 of 150. Together with the probe exemption in #249 this took the interop runner's `handshakeloss`/`handshakecorruption` cases from hard failures to passing.

Full eunit passes.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -4554,6 +4554,15 @@ process_frame(_Level, ping, State) ->
     State;
 process_frame(Level, {crypto, Offset, Data}, State) ->
     buffer_crypto_data(Level, Offset, Data, State);
+process_frame(Level, {ack, _Ranges, _AckDelay, _ECN}, State) when Level =/= app ->
+    %% Initial/Handshake ACKs must never touch the loss tracker: only
+    %% 1-RTT packets are registered there, and packet numbers restart
+    %% per space, so a Handshake-space ACK of PN 0..N silently "acks"
+    %% the first N 1-RTT packets out of sent_q. If those carried data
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks, this is a good catch and it turned out to be more serious than the description suggests.

I built a repro on the socket adapter: blackhole the client for 1.5s so a full window is stranded, then restore the path. On main the transfer never recovers, 0 of 262144 bytes in 20s, with the client re-sending everything (225 packets, 276796 bytes measured at the bridge) and the server echoing nothing. With this change it completes in ~250-550ms. So it is silent data loss on any path that drops a window and comes back: WiFi-to-cellular handover, VPN reconnect, NAT rebind.

Took your commit into #256 with the tests it was missing.

Will close this once #256 lands.

jbevemyr:
Closing in favour of #256, which carries this commit with the tests it was missing. Thanks for adding them.

PR #251, opened 2026-08-26
Keep the sent-packet tracker across an active path migration

jbevemyr:
Active migration replaced the loss tracker with `quic_loss:new()`, orphaning every in-flight packet: no ACK can match them, loss detection never examines them, and PTO never fires for them since `bytes_in_flight` reads 0. Any lost data in that set was never retransmitted, and the peer stalled on a permanent stream hole until idle timeout.

RFC 9002 §9.4 resets congestion state on a path change, not the sent-packet tracking. This PR keeps sent_q and its byte accounting, resets the RTT estimate and PTO backoff (they belong to the old path), carries the outstanding bytes into the fresh congestion controller, and marks the PTO timer dirty so it re-arms for the surviving packets.

Reproducer: a 5 MiB transfer through a proxy that rebinds the client's source address and port every second (dead old bindings, as a NAT would). Previously the transfer died on idle timeout partway in; with the fix it completes, and the interop runner's `rebind-addr` case went from failing to passing.

Full eunit and dialyzer pass.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -10739,9 +10739,18 @@ complete_migration(
     NewPMTUState = quic_pmtu:on_path_change(PMTUState),
 
     %% RFC 9002 Section 9.4: Reset congestion controller on path change
-    %% The new path may have different RTT and bandwidth characteristics
-    NewCCState = quic_cc:new(#{}),
-    NewLossState = quic_loss:new(),
+    %% The new path may have different RTT and bandwidth characteristics.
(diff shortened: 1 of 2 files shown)


benoitc:
Confirmed and carried into #258 with tests, authorship intact. Grouped with the other `quic_loss` state-lifetime fix because their tests share a harness. Will close this once #258 lands.

benoitc:
Duplicate of #258, which carries this work with tests. Closing in favour of it.

PR #252, opened 2026-08-26
Retransmit the client's Finished until the handshake is confirmed

jbevemyr:
Once the client statem leaves the handshaking state, nothing retransmitted the Finished: handshake-space packets are not in the loss tracker, and the handshake-retransmit state timeout only runs in the handshaking state. A client whose Finished was lost considered itself connected and sent 1-RTT data the server cannot act on before handshake completion, while the server replayed its flight against ACK-only responses; the connection idled out with the Finished forever unacknowledged.

This PR retains the Certificate(+CertificateVerify)+Finished payload until HANDSHAKE_DONE and resends it (CRYPTO offsets make this idempotent) in two situations:

- when a handshake-level CRYPTO duplicate arrives from offset 0, the unmistakable sign that the server's retransmission timer fired because the Finished never arrived, and
- from the PTO, per RFC 9002 section 6.2.1: until the handshake is confirmed the PTO probes the handshake space too.

Found via the interop runner's handshakeloss case, where a few of the 50 handshakes per run wedged this way under 30% loss. With this plus #249 and #250 the case passes in both directions against ngtcp2. Full eunit passes.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -362,6 +362,10 @@
     initial_tx_off = 0 :: non_neg_integer(),
     %% Client-side: CH1 random + build opts, needed to rebuild CH2
     tls_ch1_random :: binary() | undefined,
+    %% Client-side: the Certificate(+CertificateVerify)+Finished payload,
+    %% retained until HANDSHAKE_DONE so a lost Finished can be resent -
+    %% nothing else retransmits it once the statem has left `handshaking'.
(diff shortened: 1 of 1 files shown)


benoitc:

| | two dropped |
|---|---|
| main | fails |
| #252 alone | fails |
| #230 alone | fails |
| #252 + #230 | passes |

benoitc:
Duplicate of #259, which carries this work with tests. Closing in favour of it.

PR #253, opened 2026-08-26
Faster handshake retransmission cadence

jbevemyr:
The handshake flight retransmit ran at a fixed 500 ms base backoff with 8 attempts, giving up after roughly 23 seconds. On low-RTT paths under loss that is 5 to 15 times slower than PTO cadence; in the interop runner's handshakecorruption case it meant ~15 seconds per handshake, and only 19 of 50 handshakes completed within the case's budget. A few percent of handshakes under 30% loss failed outright by exhausting the attempts.

200 ms base with 12 attempts (capped at 3000 ms) tracks lossy handshakes much more closely while still backing off exponentially, and the total effort stays bounded to roughly half a minute before the connect timeout owns the outcome. With this change handshakecorruption went from 19 of 50 to all 50 within budget, and together with #249, #250 and #252 both handshakeloss and handshakecorruption pass.

Full eunit passes.

### DIFF
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -214,9 +214,12 @@
 %% so it can flush the deferred flight. Backoff doubles from the base up
 %% to the cap. Localhost handshakes complete well under the base, so the
 %% timer is cancelled by the state change and never fires.
--define(HS_RTX_BASE_MS, 500).
--define(HS_RTX_MAX_MS, 4000).
--define(HS_RTX_MAX_ATTEMPTS, 8).
+%% 200 ms base keeps a lossy handshake moving at something closer to
+%% PTO cadence on low-RTT paths; 12 attempts bound the total effort to
+%% roughly half a minute before the connect timeout owns the outcome.
+-define(HS_RTX_BASE_MS, 200).
+-define(HS_RTX_MAX_MS, 3000).
+-define(HS_RTX_MAX_ATTEMPTS, 12).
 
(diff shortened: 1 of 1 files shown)


benoitc:
Tried to build a local test for this and could not, which I think is worth recording rather than leaving as a silent gap.

I drove handshakes over a socket adapter while dropping a deterministic fraction of the client's Initial and Handshake datagrams, up to one in two, and measured connect time on main against this branch. Both complete in 2-3 ms in every configuration. On loopback the handshake finishes far below even the 200 ms base, so the retransmit timer is cancelled by the state change and never fires. The comment above the constants already says this, and it turns out to rule out a local regression test too: any assertion that separates the two cadences would have to be timing-based and would be flaky in CI.

So the evidence for this change is the interop runner (handshakecorruption 19/50 to 50/50), and that is the right place for it. I am not going to add a local test that pretends to cover it.

No objection to the numbers themselves: 12 attempts at a 200 ms base with a 3000 ms cap is about 27 s of total effort against roughly 23 s before, so the connect timeout still owns the outcome and the change is a re-shaping of the cadence rather than a longer wait.

Leaving this open for you to merge on the interop evidence. Not superseding it, since there is nothing for me to add.

jbevemyr:
Agreed on all points. The interop runner is where the cadence is observable, and handshakecorruption 19/50 to 50/50 is the evidence I trust too. Thanks for writing down why a local test cannot work here; that saves the next person the same detour.

benoitc:

PR #254, opened 2026-08-26
Cap the PTO backoff at 5 seconds

jbevemyr:
The PTO doubled without bound, so after a loss streak of eight or nine expirations the probe interval reached tens of seconds; a probe that arrives after the peer's request timeout is a probe that never happened. Measured under a 30%-loss UDP proxy with 1 KiB request/response exchanges: roughly 1% of exchanges stalled because the responder's PTO backed off past the requester's patience while the response sat unacked in sent_q (observed pto_count 8-9, meaning 10-60+ second probe intervals).

Probes are tiny, so a bounded worst-case interval costs nothing while keeping recovery inside real-world timeouts. RFC 9002 requires exponential backoff but does not forbid a ceiling, and most production stacks cap it. With the cap, 299 of 300 exchanges complete; without it, stalls reproduced in every run of 200.

Full eunit passes.

### DIFF
--- a/src/quic_loss.erl
+++ b/src/quic_loss.erl
@@ -71,6 +71,7 @@
 % 9/8
 -define(TIME_THRESHOLD, 1.125).
 % 1 millisecond
+-define(MAX_PTO_MS, 5000).
 -define(GRANULARITY, 1).
 % RFC 9002 default is 333ms, but 100ms is more aggressive for faster ramp-up
 -define(DEFAULT_INITIAL_RTT, 100).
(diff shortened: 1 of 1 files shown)


benoitc:
Confirmed and carried into #258 with tests, authorship intact. Grouped with the other `quic_loss` state-lifetime fix because their tests share a harness. Will close this once #258 lands.

benoitc:
Duplicate of #258, which carries this work with tests. Closing in favour of it.

PR #256, opened 2026-08-26
Do not apply Initial/Handshake ACKs to the 1-RTT loss tracker

benoitc:
Carries @jbevemyr's commit from #250 with the tests it was missing. Supersedes #250.

## The bug

Packet numbers restart per space (RFC 9000 §12.3) and only 1-RTT packets are registered in the loss tracker, so a Handshake-space ACK of packet numbers 0..N retired the first N 1-RTT packets from `sent_q` without the peer ever having received them. Nothing retransmits them and the peer keeps a permanent hole in the stream.

It needs a real outage to show, which is why it survived: on a healthy path the 1-RTT packets are acknowledged before any late handshake-level ACK can retire them.

Measured on main, over the socket adapter, blackholing the client for 1.5s so a full window is stranded and then restoring the path:

- the client re-sends the whole payload after the path returns (225 packets, 276796 bytes at the bridge)
- the server echoes **0 of 262144 bytes in 20s**
- final state: `bytes_in_flight => 0`, `congested => false`, send queue empty, connection still up

The data is gone and the transfer never completes. With the fix it completes in ~250-550ms. Real paths that do this: WiFi-to-cellular handover, VPN reconnect, NAT rebind.

The new clause follows the `Level =/= app` idiom already used for `handshake_done`, `max_data`, `max_stream_data`, `max_streams` and `data_blocked`.

## Tests

`quic_handshake_ack_isolation_tests` pins the fix at the frame boundary. 4 of its 9 cases fail without it, including a handshake ACK that empties the tracker outright. The application-level cases are the inverse check: they fail if the fix ever turns into "stop processing ACKs".

`quic_stranded_window_recovery_SUITE` covers the consequence end to end, with a control case running the same transfer with no outage so a failure separates a real stall from a broken harness.

`test_state_with_loss/1` and `test_loss_state/1` join the existing `-ifdef(TEST)` export block.

## Checks

`eunit` 2271/0, `dialyzer`, `xref`, `lint`, `fmt --check` all clean. `ct` shows no new failures: the QPACK interop, `quic_h3_owner_SUITE` init and `quic_psk_e2e_SUITE` failures are all present on main too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- ACKs arriving at the Initial or Handshake encryption level no longer
+  reach the 1-RTT loss tracker. Packet numbers restart per space
+  (RFC 9000 §12.3), so a Handshake-space ACK of packet numbers 0..N was
+  retiring the first N 1-RTT packets from the sent queue without the peer
+  having received them: nothing retransmitted them and the peer kept a
+  permanent hole in the stream. A path that drops a full window and then
+  returns (WiFi-to-cellular handover, VPN reconnect, NAT rebind) left the
+  transfer stalled for good. Contributed by jbevemyr (#250).
+
 ## [1.8.1] - 2026-08-15
 
 ### Added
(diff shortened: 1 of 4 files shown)


benoitc:

PR #257, opened 2026-08-27
Accept all supported groups by default on the server

benoitc:
Carries @jbevemyr's commit from #246 with the tests it was missing. Supersedes #246.

## The issue

The server's default group preference held only x25519, so a client whose key_share leads with another curve got a HelloRetryRequest even though the crypto layer speaks secp256r1 and secp384r1 fine. The handshake still completes, which is why nothing looked broken, but it costs a full extra round trip on every such connection. picoquic shares P-256 first, so that was every connection from it.

Measured over the socket adapter, against a default server:

| client `groups` | before | after |
|---|---|---|
| `[secp256r1, x25519]` | x25519, 5 client Initial datagrams | secp256r1, 3 |
| `[secp384r1, x25519]` | x25519, 5 | secp384r1, 3 |
| `[x25519]` | x25519, 3 | x25519, 3 |

## Tests

`negotiated_group` alone does not show the cost, since both outcomes connect. The suite also counts Initial-level datagrams the client sends before the connection is up. A client pads every datagram carrying an Initial to 1200 bytes while it is unvalidated (RFC 9000 §14.1), so the count reads directly on how many Initial-level exchanges the handshake needed, and an HRR shows up as more of them.

Counts are asserted against a same-run x25519 baseline rather than fixed numbers, so the cases survive unrelated changes to the handshake shape.

Two of the five cases fail without the change. The other three hold either way and are fences: x25519 clients keep their direct acceptance, an explicit `groups` option still restricts and still forces the HRR, and a disjoint client is still rejected. The HRR fence also proves the counter can see an extra round trip, so the equalities are not vacuous.

`quic_hrr_e2e_SUITE` still passes unchanged, which is the check that broadening the default did not quietly disable the HRR path.

## Checks

`eunit` 2262/0, `dialyzer`, `xref`, `lint`, `fmt --check` clean. `ct` shows no new failures against main's baseline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- A server with no explicit `groups` option now offers every classical
+  group the crypto layer supports (x25519, secp256r1, secp384r1) instead
+  of x25519 alone. A preference list holding only x25519 sent a
+  HelloRetryRequest to any client whose key_share led with another
+  curve, costing a full extra round trip on every such connection and
+  exercising the HRR path in flows that did not need it. picoquic shares
+  P-256 first, so this was every connection from it. An explicit
+  `groups` option still restricts the set exactly as before.
+  Contributed by jbevemyr (#246).
+
 ## [1.8.1] - 2026-08-15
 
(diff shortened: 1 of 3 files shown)


benoitc:

PR #258, opened 2026-08-27
Loss tracker: survive path migration, cap the PTO backoff

benoitc:
Carries two of @jbevemyr's commits with the tests they were missing. Supersedes #251 and #254.

Grouped because both are `quic_loss` state-lifetime bugs and their tests share a harness. Each is a separate commit with his authorship intact.

## Sent-packet tracker across a migration (#251)

The tracker was replaced wholesale on an active path migration, which orphans every packet already in flight: no ACK matches them, loss detection never runs, and `bytes_in_flight` reads zero so no PTO fires either. Their data is never retransmitted and the peer keeps a permanent hole in the stream.

`reset_for_new_path/1` draws the line where it belongs: the sent queue and its accounting survive, the path-derived estimates (RTT, PTO count, first-sample flag) reset, because those did belong to the old path.

This is the same failure class as #250, which I reproduced end to end this week: once packets are silently retired from the sent queue, nothing brings them back and the transfer is dead without the connection closing.

## PTO backoff cap (#254)

RFC 9002 §6.2.1 doubles the PTO on each consecutive expiration and the doubling had no ceiling. Measured on a 50 ms path:

| expirations | before | after |
|---|---|---|
| 4 | 2800 ms | 2800 ms |
| 9 | 89600 ms | 5000 ms |
| 12 | 716800 ms | 5000 ms |

Nearly twelve minutes. A probe scheduled that far out never happens: the idle timer, and any request deadline above it, fired long before. Probes are one small packet, so bounding the worst case costs nothing.

## Tests

`quic_loss_path_migration_tests`, 8 cases, all 8 fail without the change since the function does not exist.

`quic_pto_backoff_cap_tests`, 8 cases, 4 fail on main. The other 4 are fences: the backoff still doubles exactly below the cap, stays monotonic, does not clamp a fast path up to the ceiling, and leaves an unexpired PTO alone. Those matter here because a cap is easy to implement in a way that also flattens the region below it.

## Checks

`eunit` 2278/0, `dialyzer`, `xref`, `lint`, `fmt --check` clean. `ct` shows no new failures against main's baseline; `quic_server_batching_SUITE` and `quic_sigalg_e2e_SUITE` report `error_in_suite` only in full runs and pass in isolation on main too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- The sent-packet tracker now survives an active path migration. It was
+  replaced wholesale, which orphaned every packet already in flight: no
+  ACK matched them, loss detection never ran, and `bytes_in_flight` read
+  zero so no PTO fired either. Their data was never retransmitted and
+  the peer kept a permanent hole in the stream. The path-derived
+  estimates (RTT, PTO count) still reset, since those belong to the old
+  path. Contributed by jbevemyr (#251).
+- The PTO backoff is capped at 5 seconds. RFC 9002 section 6.2.1 doubles
+  the PTO on each consecutive expiration and the doubling had no
+  ceiling, so on a 50 ms path it reached roughly 90 seconds after nine
+  expirations and about twelve minutes after twelve. A probe scheduled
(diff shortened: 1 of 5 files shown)


benoitc:

PR #259, opened 2026-08-27
Recover a lost client Finished

benoitc:
Carries @jbevemyr's commits from #252 and #230 with the tests they were missing. Supersedes #252, #230 and #225 (#225 is contained in #230).

These two land together because **neither fixes the bug alone**. That is the substantive finding here.

## The bug

The client's Certificate(+CertificateVerify)+Finished flight goes out at the Handshake level. Once the client state machine leaves `handshaking` nothing retransmits it: handshake-space packets are not in the 1-RTT loss tracker, and the handshake retransmit timer only runs in that state.

So a client whose Finished is dropped reports itself connected and starts sending 1-RTT data the server cannot act on before the handshake completes, while the server replays its own flight against ACK-only answers. Nothing breaks the tie and the connection dies on the idle timer.

## Why both

`quic_lost_finished_SUITE` drops the client's first N Handshake-level datagrams at a bridge, then lets everything through, so recovery depends entirely on the client resending the flight.

One dropped datagram already recovers on main. **Two does not**: the client still reports connected and then no data ever flows. Measured over three runs each:

| | two dropped |
|---|---|
| main | fails |
| #252 alone | fails |
| #230 alone | fails |
| **#252 + #230** | **passes** |

#252 gets the flight resent; #230 is what lets the recovered connection actually drain its queue, by sending the part of a write that fits under the peer's flow-control limit rather than nothing. Landing either one on its own leaves the bug live, which is not visible from either PR in isolation.

The no-loss control case separates a real stall from a broken harness.

## Checks

`eunit` 2267/0, `dialyzer`, `xref`, `lint`, `fmt --check` clean. `ct` shows no new failures against main's baseline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- A client whose Finished is lost now recovers. The
+  Certificate(+CertificateVerify)+Finished flight goes out at the
+  Handshake level, and once the client state machine left `handshaking'
+  nothing retransmitted it: handshake-space packets are not in the 1-RTT
+  loss tracker and the handshake retransmit timer only runs in that
+  state. The client considered itself connected and sent 1-RTT data the
+  server could not act on before handshake completion, while the server
+  replayed its own flight against ACK-only answers, until the connection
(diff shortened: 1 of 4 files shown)


benoitc:

PR #260, opened 2026-08-27
Loss time threshold: max(smoothed_rtt, latest_rtt)

benoitc:
Carries @jbevemyr's commit from #210 with the tests it was missing. Supersedes #210.

## The bug

RFC 9002 §6.1.2 sets the loss time threshold from `max(smoothed_rtt, latest_rtt)`. Using the smoothed estimate alone makes the threshold too small exactly when it matters: an RTT spike outruns the EWMA, the two diverge, and in-flight packets are declared lost while their ACKs are merely late. Each spurious loss both retransmits data the peer already has and collapses the congestion window, so one latency excursion (receiver queueing, bufferbloat) becomes a throughput collapse.

This is demonstrated rather than argued from the RFC. Three packets, two sent 20 ms before the third, then only the third acknowledged with a 400 ms sample. The earlier two are 420 ms old: past `1.125 * smoothed_rtt` (75 ms) but inside `1.125 * latest_rtt` (450 ms), so the two thresholds disagree. On main they come back in the lost list. With the change they do not.

The packet-number threshold is deliberately kept out of it: both are within 3 of the acked packet number (RFC 9002 §6.1.1), so nothing else can declare them lost and the time threshold is the only thing deciding. Getting that wrong is what made my first attempt at this test misleading.

## Tests

9 cases, 2 fail on main. The rest hold either way and are fences, including the inverse case where a genuinely old flight must still be declared lost, so taking the larger RTT cannot quietly switch loss detection off.

The threshold is read by subtraction from `get_loss_time_and_space/1` against a pinned send time, so the assertions never re-derive the formula under test.

## Checks

`eunit` 2271/0, `dialyzer`, `xref`, `lint`, `fmt --check` clean. `ct` shows no new failures; the `error_in_suite` entries in full runs (`hrr_e2e`, `lb_e2e`, `sigalg_e2e`, `psk_e2e`, `server_batching`) all pass in isolation on this branch, 3/21/5/8/3 respectively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- The loss time threshold is `max(smoothed_rtt, latest_rtt)`
+  (RFC 9002 section 6.1.2) rather than the smoothed estimate alone. When
+  an RTT spike outruns the EWMA the two diverge, and the smaller
+  threshold declares in-flight packets lost while their ACKs are merely
+  late. Each spurious loss both retransmits data the peer already has
+  and collapses the congestion window, so a single latency excursion
+  (receiver queueing, bufferbloat) turned into a throughput collapse.
+  Contributed by jbevemyr (#210).
+
 ## [1.8.1] - 2026-08-15
 
 ### Added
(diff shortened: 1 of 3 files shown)


benoitc:

PR #261, opened 2026-08-27
Bind the socket-backend client to the requested source address

jbevemyr:
## The bug

The gen_udp client path passes `extra_socket_opts` (including `{ip, Addr}`) through to `gen_udp:open`, so a caller on a multi-address host can pick which address its connections originate from. The socket backend dropped those options entirely: the socket was never bound, the kernel picked the source address per route (127.0.0.1 for any loopback destination), and the peer attributed the connection to the wrong host.

We hit this in a test rig where several simulated hosts share one machine and are told apart by loopback source address; every socket-backend client appeared as 127.0.0.1.

## The fix

`configure_send_socket/3` now honors `{ip, Addr}` from `extra_socket_opts` with an explicit `socket:bind` (port 0, family derived from the address), matching the gen_udp path. A failed bind logs a warning and falls back to the previous unbound behaviour rather than failing the connect, since the address may be legitimately absent on the host (e.g. DHCP renumbering between config and connect).

### DIFF
--- a/src/quic_socket.erl
+++ b/src/quic_socket.erl
@@ -811,6 +811,7 @@ open_send_socket_backend(Family, Opts, BatchConfig) ->
 configure_send_socket(Socket, Opts, BatchConfig) ->
     ok = socket:setopt(Socket, {socket, reuseaddr}, true),
     set_socket_buffer_sizes(Socket, Opts),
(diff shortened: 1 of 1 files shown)


benoitc:

PR #262, opened 2026-08-27
GRO/GSO batching: read the cmsg, size the buffer, run mixed batches

benoitc:
Carries five of @jbevemyr's commits with the tests they were missing. Supersedes #204, #215, #217, #219 and #203.

Grouped because they are all the socket batching path and their tests share a harness. Each is a separate commit with his authorship intact. #217 already carried his own linter follow-up, so the Elvis failure on that PR is fixed here rather than needing anything from me.

## The sharp one: #204

```erlang
extract_gro_segment_size([#{level := udp, type := ?UDP_GRO, data := <<Size:16/native>>} | _])
```

The kernel's `UDP_GRO` payload is an int. A pattern accepting exactly two bytes **never matches**, so the size was never found, the coalesced buffer went up unsplit as one oversized datagram, and the QUIC layer dropped it because short-header packets cannot be re-split. GRO was silently losing every coalesced train it received.

## #215

The GRO read passed `0` for the buffer size, which uses the OTP default 8 KiB and makes `recvmsg` silently truncate any larger train. The client receiver also used a plain `recvfrom`, which never split trains at all.

## #217

A mixed-size batch fell back to one `sendmsg` per packet, so a single ACK between data packets degraded the whole batch. It now goes out as runs of equal-sized packets, one `UDP_SEGMENT` call each.

## Tests

GRO and GSO need a Linux kernel end to end, but the parts that decide what happens are pure, so they are tested directly. A TEST-only export block is added to `quic_socket` for the three, following the convention already in `quic_connection` and `quic_listener`.

3 of the 16 fail on main, all of them the cmsg read. The rest are fences, two worth calling out:

- every emitted GSO run is legal: all segments but the last exactly the segment size, which is what `UDP_SEGMENT` requires
- whatever the grouping, every packet goes out exactly once and in the order it was queued

`quic_pacing_burst_tests` covers #219, 3 of 7 failing on main. It reads the burst ceiling by letting the token bucket saturate and asking for more than it can hold, polling until two consecutive reads agree rather than sleeping a fixed interval, which keeps it honest at low rates where the bucket fills slowly.

## Checks

`eunit` 2285/0, `dialyzer`, `xref`, `lint`, `fmt --check` clean. `ct` shows no new failures; `quic_server_batching_SUITE`, `quic_adapter_SUITE` and `quic_client_socket_backend_tests` all pass in isolation on this branch (3, 4, 4).

Note the GSO send path itself is only exercised by the `Server Batching + Linux GSO` CI job, since I cannot run it on macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- The UDP_GRO control message is read as the int the kernel sends.
+  Matching exactly two bytes meant the lookup never succeeded, so a
+  GRO-coalesced buffer was passed up unsplit as one oversized datagram
+  and dropped by the QUIC layer, which cannot re-split short-header
+  packets. GRO was therefore silently losing every coalesced train.
+  Contributed by jbevemyr (#204).
+- The GRO receive path sizes its read buffer for a maximally coalesced
+  train (64 KiB). Passing 0 used the OTP default 8 KiB buffer and
+  `recvmsg' silently truncated anything larger, discarding every segment
+  past the first few. The client receiver also went through a plain
+  `recvfrom', which never split trains at all. Contributed by jbevemyr
+  (#215).
+
+### Changed
+- A mixed-size send batch on a GSO socket is split into runs of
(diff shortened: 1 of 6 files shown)


benoitc:

PR #263, opened 2026-08-27
Fix amp-budget starvation on batched receive; retransmit the server handshake flight

jbevemyr:
Two handshake-reliability fixes found with a handshake storm harness (waves of 50-150 concurrent connects through a lossy path): 0.1-0.5% of handshakes hung until the connect timeout, singly and in bursts.

## 1. Anti-amplification accounting skipped on the batched receive path

The amp-budget credit (RFC 9000 §8.1) only ran in `handle_packet/2`. The batched `{quic_packets, ...}` delivery the listener uses for GRO trains skipped it entirely, so a server whose ClientHello arrived in a batch kept `amp_rx` at or near zero and deferred most or all of its handshake flight. The client saw only budget-permitted fragments, its Initial retransmits often arrived batched again (still uncounted), and the handshake wedged permanently.

`handle_packets_batch/2` now runs the same per-datagram accounting before processing and flushes any deferred flight after the batch, exactly like the single-datagram path.

Pinned by `quic_amp_batch_accounting_tests`: every datagram in a batch must be credited for an unvalidated server, none for a validated one, and the batch itself must not touch the spend side.

## 2. A lost server handshake flight was never retransmitted

Initial/Handshake packets are not loss-tracked (`on_packet_sent` runs only on 1-RTT paths), so a genuinely lost ServerHello-to-Finished flight was gone for good: the server's TLS state had advanced, making the client's Initial retransmits produce only ACKs. Permanent wedge until the connect timeout.

The server now retains `{ServerHello, its Initial CRYPTO offset, Handshake payload}` until the client's Finished arrives, and replays the flight on the same backoff schedule the client uses for its Initial (RFC 9002 §6.2 behaviour without full per-space loss tracking). The replay uses new packet numbers over identical crypto-stream bytes, so client-side reassembly is byte-exact regardless of what already got through.

Pinned by `quic_server_flight_retransmit_SUITE`: the harness drops everything the server sends for a fixed 700 ms window (covering the original flight and the first two replay attempts), then heals the path. Without the fix the case fails on a connect timeout, with it the handshake completes on the first replay after the window and a small echo confirms working application keys. A control case fences the harness. The drop is by time window rather than packet count, so however many flights the client's Initial retransmits elicit inside the window, the case stays deterministic.

## Storm evidence

On the branch this was developed on: pre-fix roughly one failed round in 30-40 (single hangs plus waves of 12-50 when bursts hit the batched path); post-fix 200 rounds of 150 connects and 400 rounds of 50 connects clean, 50,000 handshakes, zero failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ All notable changes to this project will be documented in this file.
   (receiver queueing, bufferbloat) turned into a throughput collapse.
   Contributed by jbevemyr (#210).
 - PMTU probes are tracked as non-ack-eliciting, so a probe lost past the path MTU no longer inflates `bytes_in_flight`, arms the PTO machinery, or feeds a congestion event (RFC 8899 §3, RFC 9000 §14.4). Combined with an in-flight-keyed liveness check, the periodic raise probe previously killed every long-lived connection on an MTU-limited path once per 600-second raise interval, both ends at once. The raise interval is configurable as `pmtu_raise_interval`. (#264)
+- Anti-amplification accounting (RFC 9000 §8.1) now also runs on the batched listener delivery path. A server whose ClientHello arrived in a GRO batch kept its amp budget at zero, deferred the handshake flight, and the handshake wedged until the connect timeout. (#263)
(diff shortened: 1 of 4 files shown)


benoitc:

PR #264, opened 2026-08-28
PMTU probes are not connection loss (RFC 8899 §3, RFC 9000 §14.4)

jbevemyr:
## The bug

RFC 8899 §3 and RFC 9000 §14.4: a PMTU probe sent past the path MTU is lost **by design**, and its fate belongs to the PMTUD state machine alone. eq sent probes through `send_app_packet`, which tracked them as ordinary ack-eliciting data. A lost probe therefore inflated `bytes_in_flight`, armed the PTO machinery, and fed a spurious congestion event when loss detection retired it.

The lethal variant appears the moment a liveness check keyed on unacknowledged in-flight data exists (our deployment carries a msquic-style 16 s disconnect timeout, and #224 brings one here): the 600-second raise timer re-probes above the settled MTU, the probe is dropped by the path, `bytes_in_flight` sits above zero with no possible progress, and 16 seconds later a perfectly healthy peer is declared dead. **Every long-lived connection on an MTU-limited path — that is, most real WAN paths — died and reconnected once per raise interval, both ends simultaneously.**

We found it on a virtualized rig with a 1463-byte path: connections lived exactly `connect + 600 s + 16 s`, in a permanent cycle, with `in_flight = 1467` (the raise probe, 1463 + the 4-byte search step) in both endpoints' death logs. It never shows on loopback test rigs: with no MTU limit the search completes at the ceiling and the raise probe never exceeds the path.

## The fix

- `send_pmtu_probe_packet` sends the probe with an **empty frames list**, so it is tracked as non-ack-eliciting: it stays in the sent queue, which is what the PMTU ack/lost hooks fold over, but contributes nothing to `bytes_in_flight`, `acked_bytes`, or `lost_bytes`, arms no PTO, and is never retransmitted (the PMTUD probe timer owns retries). The acked/lost/RTT accounting already filters on the ack-eliciting flag, so the probe becomes fully inert to the connection while remaining visible to PMTUD.
- The congestion-event timestamp in `detect_lost_q` is likewise gated on ack-eliciting packets, so a lost probe cannot trigger `on_congestion_event`.
- `pmtu_raise_interval` is now read from the connect/listen options, matching the `pmtu_opts()` type that already declared it. Tests use it to run several raise cycles in seconds; deployments on known-constrained paths can tune it.
- `send_app_packet/2` lost its only caller and is removed.

## Tests

- `quic_pmtu_probe_loss_tests` pins the rule at the loss layer: a non-ack-eliciting probe adds no in-flight bytes; acked, it appears in the acked fold (for the PMTU hook) with `acked_bytes = 0` and `has_ack_eliciting = false`; lost, it appears in the lost fold with `lost_bytes = 0` and **no congestion-event timestamp** (red before the fix); a fence case keeps real data loss signaling congestion.
- `quic_pmtu_raise_probe_SUITE` runs the scenario end to end: an echo connection through a bridge that drops any datagram above 1300 bytes, a shortened raise interval, and a 20-second traffic-free window spanning several probe-and-lose cycles. The connection must stay quiet-alive throughout and still carry data afterwards. On this branch the case guards against the non-lethal damage; once a disconnect timeout exists (#224) it is the regression net for the lethal variant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- PMTU probes are tracked as non-ack-eliciting, so a probe lost past the path MTU no longer inflates `bytes_in_flight`, arms the PTO machinery, or feeds a congestion event (RFC 8899 §3, RFC 9000 §14.4). Combined with an in-flight-keyed liveness check, the periodic raise probe previously killed every long-lived connection on an MTU-limited path once per 600-second raise interval, both ends at once. The raise interval is configurable as `pmtu_raise_interval`. (#264)
+
 ## [1.8.1] - 2026-08-15
 
 ### Added
--- a/src/quic_connection.erl
+++ b/src/quic_connection.erl
@@ -663,6 +663,7 @@
     pmtu_state :: #pmtu_state{} | undefined,
     pmtu_probe_timer :: reference() | undefined,
     pmtu_raise_timer :: reference() | undefined,
+    pmtu_raise_interval = ?PMTU_DEFAULT_RAISE_INTERVAL :: pos_integer(),
 
     %% Deferred PTO timer reset, flushed at batch boundaries via
     %% flush_dirty_timers/1. (Idle and keep-alive timers are lazy and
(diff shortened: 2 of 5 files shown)


benoitc:

PR #265, opened 2026-08-29
The terminate-path CONNECTION_CLOSE never reached the wire

jbevemyr:
## The bug

`send_connection_close` sends the close frame through the normal packet path, which **batches** it into an updated `socket_state` — and returns that updated state to nobody. `terminate/3` called it for its side effect, then flushed the **stale** `socket_state` still in the state record. The close frame lived only in the discarded batch and never reached the wire.

Every teardown that goes through `terminate/3` — the connection's owner process dying, an exit signal from a supervisor or a controlling application — therefore closed our side **silently**. The peer kept a phantom connection until its idle timeout: up to a minute of a "connected" host whose stack had already torn the connection down. We found it on a rig where an application deliberately closes CT links by killing the process that owns them; the control tower kept reporting the site connected for the whole detection window. A clean `quic:close/2` was unaffected (that path keeps the updated state), which is exactly why this hid: the common path worked, the failure paths didn't.

## The fix

`send_connection_close` returns the updated state, and `terminate/3` flushes and closes **that** `socket_state`. Draining/closed states keep the old behavior (close already sent).

## Test

`quic_owner_down_close_SUITE`: server-side connections owned by the test process, client owned by a disposable process. Kill the owner → the peer must see `{closed, _}` within 2 s (far below any idle/disconnect timeout it would otherwise wait out). A clean-close fence separates owner-death failures from harness problems. Red before the fix with `{peer_never_saw_close, within_ms, 2000}`, green after.


benoitc:

PR #274, opened 2026-09-02
Receive pass: drain, one ACK per pass, ordered reassembly, lean fast paths

jbevemyr:
First of the performance topics from our production run (volga, a pub/sub transport that moves bulk data over eq between sites). This one is the receive pass; the others (crypto NIF fusion on top of #216, opt-in delivery coalescing, the send side, an in-node profiler) follow once this has landed, each rebased on the previous. Every commit stands alone with its own test and CHANGELOG entry; the order is the dependency order.

## What changes

1. **Drain the mailbox into one receive pass; bound the socket-backend receive queue.** In connected state, after one datagram message, further datagram messages already queued (up to 64) are handled in the same pass and the ACK, socket-batch and timer flushes run once for the lot. On the socket backend the listener and the client receiver emulate a bounded kernel rcvbuf: trains go out in chunks of at most 8 packets and are tail-dropped (head packet kept, so the peer still gets a timely dup-ACK) once the connection mailbox is over 32 messages. An unbounded mailbox turned receiver overload into queueing delay instead of loss, and the peer's RTT estimator and loss detector could not tell the difference.
2. **One ACK per receive pass.** Count-based decimation only counts while a pass is active; `finish_recv_pass/1` emits one ACK for the whole train. A 64-packet pass used to emit ~32 ACKs, each a packet build, AEAD seal and send here and a decrypt plus ACK-frame pass at the sender, and the sender's window was released in 2-3 packet quanta, keeping GSO batches near size 1. The max_ack_delay timer and the reordering immediate-ACK path are unchanged.
3. **Ordered reassembly buffers.** Stream and CRYPTO reassembly buffers are `gb_trees`. A miss at the delivery point (once per received packet while a hole is outstanding) is one smallest-key lookup; the trim for repacketized overlaps walks only the keys below the point. The map version rebuilt the whole buffer on every miss, which with a multi-megabyte hole was 40% of connection CPU. Semantics as in #269: chunks ending at or before the point are dropped, straddling chunks re-keyed, longest chunk per offset wins.
4. **Fused per-packet 1-RTT receive bookkeeping.** PN space, spin bit and activity stamp in one `#state{}` rebuild instead of three (~10% of receive CPU).
5. **Lean receive fast paths.** A lean first clause for the dominant stream-frame shape (existing stream, in-order, empty buffer, no FIN, no reset, both windows comfortably open) that is one stream-record update and one map put, falling back to the general path for everything else; a sequential PN skips the ACK-range cap scan; a packet led by a stream frame skips the datagram-only delayed-ACK check.

## Numbers

Interleaved A/B on this machine, `quic_throughput_bench:run_download_sink/1` with a 10 MB download, three rounds each, main = f33f356:

| backend | main (MB/s) | this PR (MB/s) |
|---|---|---|
| gen_udp | 16.6 / 18.5 / 16.6 | 24.6 / 24.8 / 25.0 |
| socket | 20.1 / 20.8 / 20.8 | 26.7 / 33.5 / 36.1 |

The per-commit profiles in the commit messages come from the production rig (200 MiB single-stream bulk over loopback and over a 1% loss link), where the receive pass took the socket backend from 15 to 30 MB/s with retransmits down ~5x and the ordered buffer removed the loss-recovery CPU cliff.

## Tests

- `quic_recv_drain_tests`: the drain takes only datagram messages, in order, up to the cap, never once a close has been initiated.
- `quic_recv_queue_bound_tests`: chunking and tail-drop on both socket-backend paths.
- `quic_recv_pass_ack_tests`: no flush inside a pass, one ACK at its end, below-tolerance remainders keep their timer, no ACK when the pass started a close, decimation resumes afterwards.
- `quic_reassembly_tests`: converted to the tree API, plus two cases pinning the walk bound.
- `quic_record_app_recv_tests`: the fused update against the three helpers over random packet sequences, both roles, spin on and off.
- `quic_stream_recv_fast_path_tests`: lean stream-data clause against the general path (identical state and owner messages for in-order runs and for every shape it must hand over), the PN-space update against the always-cap reference over sequences that overflow the range cap, the delayed-ACK short-circuit against the general clause.

Full eunit, dialyzer, lint, and the batching, stranded-window, owner-down and handshake-flight CT suites pass.

## Left out on purpose

The production branch also carried a global loss-reason counter and an amp-fold skip in `handle_packets_batch`; the first is bench diagnostics in the loss hot path and the second was A/B-neutral with a flush edge case, so neither is here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,14 @@ All notable changes to this project will be documented in this file.
   GSO batches near size 1. The max_ack_delay timer still bounds ACK
   latency for below-tolerance remainders and the reordering
   immediate-ACK path is unchanged.
+- The out-of-order reassembly buffers (stream data and CRYPTO) are
+  ordered trees instead of maps. A miss at the delivery point, which
+  happens once per received packet while a hole is outstanding, is now
+  answered with one smallest-key lookup, and the trim walk for
+  repacketized overlaps visits only chunks below the delivery point.
+  The map version rebuilt the whole buffer on every miss; with a
+  multi-megabyte hole under loss recovery that walk dominated receiver
+  CPU.
 ### Changed
 - A server with no explicit `groups` option now offers every classical
   group the crypto layer supports (x25519, secp256r1, secp384r1) instead
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -256,6 +256,14 @@ All notable changes to this project will be documented in this file.
   bit, activity stamp) is one state update instead of three; the three
   separate helpers each rebuilt the full connection state and together
   cost about a tenth of receive CPU on bulk flows.
+- Receive-side fast paths for the dominant bulk shape: an in-order
+  stream frame on an existing stream with an empty reassembly buffer
+  and both flow-control windows comfortably open is one stream-record
+  update and one map put, falling back to the general path for every
+  other shape; a sequential packet number extends the head ACK range
+  without the range-cap scan; a packet led by a stream frame skips the
+  datagram-only delayed-ACK check. The per-frame `max_stream_data_check`
+  debug log is gone, it cost a logger allow-check per frame.
 ### Changed
 - A server with no explicit `groups` option now offers every classical
   group the crypto layer supports (x25519, secp256r1, secp384r1) instead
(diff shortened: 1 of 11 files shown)


jbevemyr:

PR #282, opened 2026-09-03
Opt-in delivery coalescing: one owner message per stream per receive pass

jbevemyr:
Second of the performance topics, on top of #274. One commit.

## What changes

With `delivery_coalescing => true`, consecutive same-stream `stream_data` deliveries within one receive pass merge into a single owner message, flushed at the end of the pass, as soon as a different stream delivers, or before a `stream_reset` notification for any stream. Only adjacent chunks of one stream merge, so the owner observes messages in exact arrival order; a first cut that coalesced per stream across the whole pass reordered control-vs-data stream delivery and broke a consumer's handshake. Owners otherwise wake once per packet on bulk flows, and that wakeup was a large share of receiver cost.

It is opt-in. QUIC makes no message-boundary guarantee, but owners that decode each delivery as one complete term rather than length-framing the byte stream break when two messages arrive in one pass; we saw that as a missing message in a production consumer under load. The default keeps the one-message-per-packet behaviour.

## Numbers

Measured on the production rig together with the run sender it was developed with (loopback, 200 MiB single stream): socket backend 59-66 to 76-81 MiB/s, gen_udp 40 to 45. Our transport runs with it on.

## Tests

`quic_delivery_coalescing_tests`: merging, FIN in the run, cross-stream flush order, a gap filled mid-pass, the default and the outside-a-pass behaviour, flushing when the pass initiated a close, and a RESET_STREAM that must not overtake data delivered before it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,14 @@ All notable changes to this project will be documented in this file.
   GSO batches near size 1. The max_ack_delay timer still bounds ACK
   latency for below-tolerance remainders and the reordering
   immediate-ACK path is unchanged.
+- The out-of-order reassembly buffers (stream data and CRYPTO) are
+  ordered trees instead of maps. A miss at the delivery point, which
+  happens once per received packet while a hole is outstanding, is now
+  answered with one smallest-key lookup, and the trim walk for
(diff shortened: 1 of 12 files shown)


jbevemyr:

PR #283, opened 2026-09-03
Send side: pacing unfreeze, chunk-run sender, listener sweep, shared sender

jbevemyr:
Third of the performance topics, on top of #274 (receive pass) and the delivery-coalescing PR: the send side, which is where the syscall time lives. Seven commits, each with its own test and CHANGELOG entry, in dependency order. Two of them are plain bugs found on the way.

## What changes

1. **Pacing unfreezes on sub-millisecond links; burst continuations leave the timer wheel.** RTT samples are whole milliseconds, so a loopback-class link reports 0 and `update_pacing_rate` skipped the update forever: the rate stayed at its handshake value while cwnd grew. And `arm_burst_continuation` used `send_after(0)`, whose ~1 ms timer-wheel tick clocked every 64-packet burst at 64 packets/ms (~85 MB/s) regardless of rate. A CRC-verified single-stream loopback transfer went from 83 to 194 MiB/s with these two.
2. **Bulk stream chunks go out as runs with one bookkeeping pass.** After the first full-size chunk is approved, as many further chunks as cwnd, pacing and the burst budget allow are approved up front, sealed and handed to the socket in one loop, and loss, CC, packet-number and counter state is updated once for the run. The approval is batched too (`quic_cc:send_check_run/4`, `on_packets_sent/2`, `quic_loss:on_packets_sent_run/3`; native for NewReno, a fold for other algorithms), and it divides the window room by the packet size so chunks already approved count against it, which repeated `send_check` calls without in-flight accounting did not do.
3. **Receive queue bound 512.** At 32 the run sender overran the receiver on loopback and burned throughput in drop/retransmit cycles (1.5 to 2.8k retransmits per 10 MB); at 512 flow control and cwnd are the backpressure.
4. **Listener receive sweep; PTO re-arm deferred to the end of the pass.** Distinct flows never GRO-coalesce, so a server with many sparse connections woke a connection per datagram. The listener drains its queued datagram messages into one sweep (256 packets), groups per source address and dispatches each group as one batch. ACK processing sets `pto_dirty` instead of re-arming per ACK frame.
5. **The send batch is flushed on entering `connected`.** The state-enter handler sends queued data, NEW_TOKEN and the first PMTU probe through the send batch and returned without flushing it, so those packets left with the next event on the connection (a delayed ACK or a PTO later). `get_stats` now reports `send_batch_pending`, which is what caught it.
6. **Server sends go through one shared sender per listener** (socket backend). Concurrent `sendmsg` on one socket handle from many connection processes burns ~20x the CPU in NIF-level contention (98.6 vs 4.7 us per send at 50 concurrent senders). Connections keep their batch buffers and GSO grouping; the sender performs the syscalls serially and owns the socket-global GSO state. A dead sender means a direct path for that connection and a fresh sender for new ones. `gso_flushes` becomes listener-wide in that mode, through a shared counter.

## Numbers

Interleaved A/B on this machine, `quic_throughput_bench:run_download_sink/1`, 10 MB, three rounds each, main = f33f356, this branch = #274 + coalescing + this PR:

| backend | main (MB/s) | this stack (MB/s) |
|---|---|---|
| gen_udp | 16.2 / 16.8 / 17.6 | 29.0 / 29.6 / 30.9 |
| socket | 20.8 / 21.0 / 21.5 | 38.2 / 38.5 / 38.8 |

On the production rig: the pacing fixes alone took verified bulk from 83 to 194 MiB/s; the shared sender took a 50-connection fan-out model from 24.3 to 11.0 s of server CPU and the 50-site test's system time from 25 to 17.7 s.

## Tests

- `quic_pacing_burst_tests` (zero-RTT case rewritten to the new semantics), `quic_burst_continuation_tests`.
- `quic_send_run_bookkeeping_tests`: the batched CC/loss primitives against their per-packet counterparts; `quic_bulk_run_send_SUITE`: 3 MiB byte-exact echo round trip on both backends, pacing on and off.
- `quic_listener_recv_sweep_tests`: drain order, budget, backend guard, per-flow grouping.
- `quic_post_handshake_flush_SUITE`: nothing pending after `connected` with a quiet peer behind a blackholed bridge (a guard on the fixed behaviour; the flake it was distilled from is the server-side baseline in `quic_server_batching_SUITE`).
- `quic_shared_sender_tests`: hand-off and counters, empty batch, dead-sender fallback, end-to-end through a real socket-backend socket.

Full eunit, dialyzer, lint, and the batching (with `QUIC_ENABLE_GSO_TEST`), bulk, handshake-flight, stranded-window, owner-down and PMTU CT suites pass.

## Left in our fork

The `sendmmsg` NIF that drains the shared sender's mailbox into one syscall (the remaining third of the 50-site CPU gap) stays out, per the #216 discussion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ All notable changes to this project will be documented in this file.
   message-boundary guarantee, but owners that decode each delivery as
   one complete application message break when deliveries merge, so it
   is opt-in.
+- Bulk stream sends go out as runs: once the first full-size chunk is
+  approved, as many further chunks as congestion control, pacing and
+  the burst budget allow are approved up front (consuming pacing
+  tokens exactly as one-at-a-time sends would), sealed and handed to
+  the socket in one loop, and loss, congestion, packet-number and
+  counter bookkeeping is updated once for the run. The per-packet
+  connection-state rebuild was the largest own-time item on the
+  bulk-send profile; a 64-packet run replaces 64 copies with one.
 - `versions` lists the QUIC versions a connection will also accept, for
   RFC 9368 compatible version negotiation. Contributed by jbevemyr (#243).
 - `hibernate_after` (default 5000 ms) hibernates an idle connection
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,14 @@ All notable changes to this project will be documented in this file.
   GSO batches near size 1. The max_ack_delay timer still bounds ACK
   latency for below-tolerance remainders and the reordering
   immediate-ACK path is unchanged.
+- The out-of-order reassembly buffers (stream data and CRYPTO) are
+  ordered trees instead of maps. A miss at the delivery point, which
+  happens once per received packet while a hole is outstanding, is now
+  answered with one smallest-key lookup, and the trim walk for
+  repacketized overlaps visits only chunks below the delivery point.
+  The map version rebuilt the whole buffer on every miss; with a
+  multi-megabyte hole under loss recovery that walk dominated receiver
+  CPU.
 ### Changed
 - A server with no explicit `groups` option now offers every classical
   group the crypto layer supports (x25519, secp256r1, secp384r1) instead
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,18 @@ All notable changes to this project will be documented in this file.
   transfer in 16 writes, against 0 once ordered). Throughput on loopback
   is unchanged; the cost is the reassembly buffer, and any real path where
   reordering matters.
+- Pacing no longer freezes on sub-millisecond links. RTT samples are
+  whole milliseconds, so such a link reports a smoothed RTT of 0 and
+  `update_pacing_rate` treated that as "no RTT yet" and skipped the
+  update: the rate stayed at its handshake-time value while cwnd kept
+  growing, clocking the connection at that stale rate forever. The RTT
+  is floored at 1 ms for the rate computation instead.
+- Burst continuations are sent to the connection directly rather than
+  through `erlang:send_after(0, ...)`. The timer wheel's ~1 ms service
+  tick turned every 64-packet burst continuation into a 64 packets per
+  millisecond clock, capping bulk streams at roughly 85 MB/s regardless
+  of the paced rate. Together with the previous fix, a verified
+  single-stream loopback bulk transfer went from 83 to 194 MiB/s.
 - An authenticated distribution connection no longer drops the peer's
   first handshake message about half the time. The `auth_callback` path
   put a short-lived gatekeeper process in front of the dist controller
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,16 @@ All notable changes to this project will be documented in this file.
   the burst budget allow are approved up front (consuming pacing
   tokens exactly as one-at-a-time sends would), sealed and handed to
   the socket in one loop, and loss, congestion, packet-number and
-  counter bookkeeping is updated once for the run. The per-packet
-  connection-state rebuild was the largest own-time item on the
-  bulk-send profile; a 64-packet run replaces 64 copies with one.
+  counter bookkeeping is updated once for the run:
+  `quic_cc:send_check_run/4` approves the run against cwnd and pacing
+  in one step (never more permissive than the sequential checks),
+  `quic_cc:on_packets_sent/2` and `quic_loss:on_packets_sent_run/3`
+  fold it into one record update each (native for NewReno, a fold for
+  other algorithms). The per-packet connection-state rebuild was the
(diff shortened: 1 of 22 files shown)


benoitc:
Thanks! However the five PRs (#274, #282, #283, #285 and the interop PR) are nested rather than independent: #282 includes #274, #283 includes #282, and #285 includes #283. This makes the review difficult because we keep reading the same commits again, and merging one PR leaves the others mostly showing already merged changes.

Could you instead put these changes into one PR, with each part kept as a separate, self-contained commit? The separation you already made looks good, and this would avoid the stacked PRs while still letting us review each change independently.

For the order, I would do: interop first, then receive, then delivery coalescing. These look good to me. The send side needs another look, and the NIF should come last.

The main concern is the send side. In my tests it is slower than `main` and introduces retransmits on a lossless loopback. 50 MB delivered and verified by byte count and CRC, socket backend, three runs each:

|      | MB/s           | packets sent     | retransmits    |
|------|----------------|------------------|----------------|
| main | 75.6 to 77.0   | 38,514           | 0              |
| #274 | 51.8 to 64.2   | 38,511 to 46,251 | 0 to 17        |
| #282 | 61.2 to 66.6   | 38,512           | 0              |
| #283 | 51.9 to 56.5   | 45,995 to 46,742 | 1,789 to 2,490 |
| #285 | 611.5 to 712.6 | 38,512           | 0              |

So around 2,000 retransmits per 50 MB with no network in the path, and roughly 20% more packets. #274 and #282 are clean, so the regression appears with the send-side changes. My first suspect is the pacing change on sub-ms links, as this looks like the receiver-overrun case the per-drain burst budget was meant to avoid.

To reproduce on `main` and then on the branch:

```sh
./certs/generate_certs.sh

docker build --build-arg OTP_VERSION=29 \
  --build-arg REBAR3_URL=https://s3.amazonaws.com/rebar3-nightly/rebar3 \
  -f docker/benchmark/Dockerfile -t quic-bench-29 .

docker run --rm -v "$PWD":/src -w /src --entrypoint sh quic-bench-29 -c '
  rebar3 compile >/dev/null && rebar3 as test compile >/dev/null
  erl -noshell -pa _build/test/lib/quic/ebin -pa _build/test/lib/quic/test -eval "
    application:ensure_all_started(quic),
    lists:foreach(fun(_) ->
      X = quic_throughput_bench:run_sink(#{data_size => 50*1048576,
                                           socket_backend => socket, port => 47000}),
      S = maps:get(client_stats, X, #{}),
      io:format(\"~7.1f MB/s  packets_sent=~p retransmits=~p status=~p~n\",
        [maps:get(mb_per_sec,X,0.0), maps:get(packets_sent,S,0),
         maps:get(retransmits,S,0), maps:get(status,X)]),
      timer:sleep(300)
    end, lists:seq(1,3)), halt(0)."'
```

`status=ok` means all bytes were received and the CRC matched. 38,512 packets is basically the floor for 50 MB, so anything above that is retransmission.

The receive side is the good part: 10 MB, socket backend, median of five gives 51.0 MB/s on `main`, 100.3 on #283 and 149.7 on #285.

The NIF results look good as well. With the shared object actually loaded, I get a median of 149.7 MB/s versus 100.3 without it, which is close to the 45% improvement you measured.

There are still two things I would keep in mind. The upload jump between #283 and #285 is probably not only coming from the crypto optimization. #283 sends around 46,000 packets with roughly 2,000 retransmits, while #285 is back to 38,512 with no retransmits. So part of that gain is also coming from avoiding the send-side regression above.

Also, loopback is probably a very favorable case for the fused path because it gives long GRO runs. I would like to compare with the numbers from your test rig before assuming we will see the same improvement in more realistic conditions.

jbevemyr:
Thanks, that was a real regression and your recipe reproduced it here on the first run (main 30-32 MB/s with 0 retransmits, #283 13-22 MB/s with 1.6-4.0k). Bisecting the six send-side commits put the start at the first one, and the retransmit count matched the listener's own tail-drop counter exactly, with zero kernel drops: the loss was the receive-queue bound from the receive-pass commits, not the network. The pacing unfreeze and the burst continuation are each clean on their own (80 and 88 MB/s, no drops); together they let the sender fill cwnd in one pass, and on loopback cwnd reaches the peer's flow-control window before any loss, so a receiver that had advertised 16 MB was dropping at a 4.9 MB backlog inside it.

Fixed by sizing the bound by the window (the connection's `max_data` in full-size packets, floored at 512 messages): 86-90 MB/s, 38.5k packets, 0 retransmits on the same runs, download unchanged. With the NIF commits on top: 306-348 MB/s, 0 retransmits, so the #285 number was not hiding a send-side gain, it just was not paying for the loss.

Everything is now one PR as you asked, #288, in the order interop, receive, coalescing, send, NIF, each commit self-contained. Closing the five. Rig numbers for the NIF path will follow on that PR.

jbevemyr:

PR #284, opened 2026-09-03
interop: declare the robustness cases, multiconnect per file, full cert chain

jbevemyr:
The interop-runner changes from our robustness campaign that had not been submitted; #245 (merged via #273) carried the http3 and resumption work, these two commits are the rest of the client and server side.

- Declare the passive robustness cases: longrtt, blackhole, amplificationlimit, handshakeloss, transferloss, handshakecorruption, transfercorruption, rebind-port, rebind-addr. The simulator induces the loss, corruption, latency or rebinding; the endpoint just transfers.
- Run `multiconnect` as one connection per file, sequentially and without retries: the runner maps handshakeloss and handshakecorruption to that testcase name and counts one handshake per requested file, so patience has to live in the handshake retransmission, not in reconnects.
- Disable `disconnect_timeout` in both endpoints: the blackhole case blacks the network out on purpose and expects the transfer to outlast it. Wait 60 s per download for the same reason.
- The server loads the full certificate chain from cert.pem, so a runner-provided chain is presented whole.

With these, our runner matrix against aioquic, quic-go and picoquic passed the declared cases; the Interop Tests job here is the check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,6 +194,14 @@ All notable changes to this project will be documented in this file.
 - A server handshake flight lost on the wire is retransmitted on the client-Initial backoff schedule until the client's Finished arrives. Initial/Handshake packets are not loss-tracked, so a lost flight previously wedged the handshake permanently: the client's Initial retransmits only elicited ACKs once the server TLS state had advanced. (#263)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


jbevemyr:

PR #285, opened 2026-09-03
Fused crypto NIF path: seal and open runs, parse frames in C, vectorize trains

jbevemyr:
The optional C path, on top of #283, as ten commits each with its own tests and CHANGELOG entry. Everything here is behind the crypto NIF: without a compiler, or with `QUIC_DISABLE_NIFS`, the library runs today's pure-Erlang code byte for byte, and the build is best effort under both rebar3 and erlang.mk.

We know the AEAD context alone is worth about 2% of connection CPU, which is what #216 was and why #275 covers most of it in Erlang. What the rest of this stack does is fuse the whole per-packet chain across a run of packets, so that Erlang does one bookkeeping pass per train instead of one per packet. That is where the doubling comes from, and it needs the context resource as a base, which is why #216 is the first five commits again.

## What changes

1. **AEAD context NIF** (#216 as it was: EVP context per key as a resource, `seal`/`open`/`hp_block`, iodata arguments, dialyzer-clean stubs, best-effort build, runtime opt-out). Rebased over #275: `quic_aead:ecb_mask/3` is the pure-Erlang fallback behind `quic_aead_ctx:hp_block/3`.
2. **`protect_run`**: a bulk chunk run is sealed in one call. Header build, nonce, AEAD and header protection for the whole run in C; AES only, ChaCha and NIF-less builds seal per packet. Send-side crypto 9% to 2% of connection CPU. Differential-tested byte for byte against `protect_short_packet` across packet-number length boundaries.
3. **`open_packet`**: a received 1-RTT packet is opened in one call (header unprotection, PN reconstruction per RFC 9000 Appendix A, AEAD open). Runs only while no key update is in flight and checks the key-phase bit before decrypting; a phase change goes to the two-stage path, which keeps the RFC 9001 §6 bookkeeping. Round-trip, stale-largest, phase-mismatch, bad-tag and ChaCha cases tested.
4. **`open_run`**: a whole train of short-header datagrams (a GRO train at the server, the drained mailbox at the client) opened in one call, the largest received PN carried forward in C, stopping at the first packet that needs the generic path. Differential-tested against sequential `open_packet`.
5. **Frame parsing in the NIF**: after the batched decrypt, `open_run` returns a frame list term-identical to `quic_frame:decode/1` for PING, ACK (with ECN), CRYPTO, STREAM, MAX_DATA, MAX_STREAM_DATA, MAX_STREAMS and HANDSHAKE_DONE, stream and crypto payloads as zero-copy sub-binaries; any other frame type returns the packet raw to the Erlang decoder, which keeps every error semantic. Differential-tested per frame type and fuzzed (random and mutated sequences, 20k iterations clean here). 50-connection server model: -8% reductions; verified bulk 190 to 222 MiB/s.
6. **Train vectorization**: a run that continues the receive sequence exactly gets one receive-bookkeeping update and one ACK-decimation increment for the whole train, and a run of one stream frame per packet, same stream, contiguous, no FIN, gets one flow-control check, one stream update, one delivery. Every deviation takes the per-packet path; tested against it shape by shape. Verified bulk with delivery coalescing: 339 to 405 MiB/s.

## Numbers

Interleaved A/B on this machine, `quic_throughput_bench:run_download_sink/1`, 10 MB, three rounds each, #283 (the pure-Erlang stack) against this PR:

| backend | #283 (MB/s) | this PR (MB/s) |
|---|---|---|
| gen_udp | 29.3 / 30.9 / 31.2 | 33.4 / 34.2 / 34.6 |
| socket | 39.8 / 40.6 / 42.7 | 58.5 / 60.4 / 61.4 |

gen_udp gains little because it hands the connection one datagram at a time, so `open_run` sees runs of one; the socket backend's GRO trains are what the fused path is for. On our production rig, delivery-verified 200 MiB to 2 GiB single-stream transfers: about 200 MiB/s with the pure-Erlang stack, about 400 with this one (msquic on the same host: 803). At 50 connections the same code took the server's CPU from 2.5x msquic to parity together with #221.

## Tests

`quic_protect_run_tests`, `quic_open_packet_tests`, `quic_open_run_tests`, `quic_nif_frame_parse_tests`, `quic_nif_fuzz` (harness, `quic_nif_fuzz:run(N)`), `quic_stream_train_tests`, plus the #216 suites. Full eunit, dialyzer, lint, and the batching, bulk, e2e, owner-down, stranded-window and PMTU CT suites pass with the NIF loaded; the same suites pass with `QUIC_DISABLE_NIFS=1`.

## Left in our fork

The `sendmmsg` NIF for the shared sender (one syscall for many destinations) is the one piece not here; it is worth another 10% of server CPU at 50 connections and can follow if this lands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ certs/*.pem
 certs/*.key
 doc/
 rebar3.crashdump
+/priv/quic_crypto_nif.so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ All notable changes to this project will be documented in this file.
   message-boundary guarantee, but owners that decode each delivery as
   one complete application message break when deliveries merge, so it
   is opt-in.
+- Bulk stream sends go out as runs: once the first full-size chunk is
+  approved, as many further chunks as congestion control, pacing and
+  the burst budget allow are approved up front (consuming pacing
+  tokens exactly as one-at-a-time sends would), sealed and handed to
+  the socket in one loop, and loss, congestion, packet-number and
+  counter bookkeeping is updated once for the run. The per-packet
+  connection-state rebuild was the largest own-time item on the
+  bulk-send profile; a 64-packet run replaces 64 copies with one.
 - `versions` lists the QUIC versions a connection will also accept, for
   RFC 9368 compatible version negotiation. Contributed by jbevemyr (#243).
 - `hibernate_after` (default 5000 ms) hibernates an idle connection
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,14 @@ All notable changes to this project will be documented in this file.
   GSO batches near size 1. The max_ack_delay timer still bounds ACK
   latency for below-tolerance remainders and the reordering
   immediate-ACK path is unchanged.
+- The out-of-order reassembly buffers (stream data and CRYPTO) are
+  ordered trees instead of maps. A miss at the delivery point, which
+  happens once per received packet while a hole is outstanding, is now
+  answered with one smallest-key lookup, and the trim walk for
+  repacketized overlaps visits only chunks below the delivery point.
+  The map version rebuilt the whole buffer on every miss; with a
+  multi-megabyte hole under loss recovery that walk dominated receiver
+  CPU.
 ### Changed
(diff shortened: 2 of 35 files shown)


jbevemyr:

PR #288, opened 2026-09-05
Performance series: interop cases, receive pass, delivery coalescing, send side, fused crypto NIF

jbevemyr:
Replaces #274, #282, #283, #284 and #285 with one series on top of 1.8.2, in the order you asked for: interop, receive pass, delivery coalescing, send side, NIF. Each commit is self-contained: it builds, carries its own tests and CHANGELOG entry, and can be reviewed on its own; the series is green (eunit, CT with and without the NIF, dialyzer, lint, erlfmt) at the tip.

## The send-side regression

Reproduced with your recipe (50 MB upload sink, socket backend, loopback, three runs):

|                        | MB/s         | packets sent    | retransmits   |
|------------------------|--------------|-----------------|---------------|
| main                   | 29.9 to 31.7 | 38,518 to 38,522 | 0             |
| #283 as pushed         | 12.7 to 22.2 | 45,621 to 48,268 | 1,583 to 4,033 |
| this series, send side | 86.1 to 90.4 | 38,522 to 38,526 | 0             |

Every retransmit matched a drop counted by the listener's own tail-drop and the kernel reported none, so the loss was ours, not the network's. Bisecting the six send-side commits put the start at the first one, and splitting that commit showed the two halves are each clean (pacing unfreeze alone 80 MB/s, burst continuation off the timer wheel alone 88 MB/s, zero retransmits either way). Together they let the sender fill its congestion window in one pass, and on loopback that window reaches the peer's flow-control window before any loss. The receiver had advertised 16 MB and was dropping at a 512-message (about 4.9 MB) mailbox backlog, inside the window it had promised.

The fix is in the commit "Size the socket-backend receive queue bound by the flow-control window" (it replaces the old "raise the bound to 512" commit): the bound is now the connection's `max_data` in full-size packets, floored at 512 messages, so a peer inside its window is never dropped and the bound only catches one outside it. Download is unchanged by it (10 MB sink: 40 MB/s against 38 before).

With the NIF path on top the same upload runs at 306 to 348 MB/s with 38,514 to 38,516 packets and zero retransmits, so the #285 number you saw was not carrying a hidden send-side gain; the fix simply removes the loss that #283 was paying for.

## Rig numbers

Measured on two physical hosts over 1 GbE (RTT 0.2 ms), see the comment below for the tables and the two findings the rig made: an ordering bug in the client mailbox drain (fixed in "Open a GRO train of short-header packets in one NIF call", with a test) and the interaction between qdisc drops and PMTU black-hole detection. Download: main 25 to 28 MB/s, series without NIF 30 to 51, series with NIF 69 to 70, zero retransmits. Upload is flat against main.

## Commits

- interop: multiconnect, robustness-case declarations, no disconnect timeout
- interop: load the full certificate chain from cert.pem
- Drain the mailbox into one receive pass; bound the socket-backend receive queue
- Flush one ACK per receive pass instead of one per tolerance window
- Ordered reassembly buffers
- Fuse the per-packet 1-RTT receive bookkeeping into one state update
- Lean receive fast paths for the bulk case
- Opt-in delivery coalescing: one owner message per stream per receive pass
- Unfreeze pacing on sub-ms links and take burst continuations off the timer wheel
- Send bulk stream chunks as runs with one bookkeeping pass
- Size the socket-backend receive queue bound by the flow-control window
- Listener receive sweep; defer the ACK-path PTO re-arm to the end of the pass
- Flush the send batch when entering connected
- Serialize server sends through a shared sender process
- Add optional AEAD-context NIF with pure-Erlang fallback
- Accept iodata in the AEAD NIF data arguments
- Give the NIF stubs specs dialyzer can use
- Make the NIF c_src build best effort under erlang.mk too
- Runtime opt-out via QUIC_DISABLE_CRYPTO_NIF / QUIC_DISABLE_NIFS
- Seal chunk runs with one fused NIF call
- Fuse the short-header receive path into one NIF call
- Open a GRO train of short-header packets in one NIF call
- Parse frames in the open_run NIF for the batched receive path
- Vectorize contiguous packet and stream trains in the batched receive path
- NIF stubs raise nif_error so dialyzer keeps their specs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ certs/*.pem
 certs/*.key
 doc/
 rebar3.crashdump
+/priv/quic_crypto_nif.so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ All notable changes to this project will be documented in this file.
   endpoints so the blackhole case can outlast its outage, waits 60 s per
   download, and the server loads the full certificate chain from
   cert.pem.
+- The connected-state receive pass drains the datagram messages already
+  queued in the connection's mailbox (up to 64) before flushing ACKs,
+  socket batches and timers, so those flushes amortize over a train
+  instead of running once per datagram. On the socket backend the
+  listener and the client receiver now emulate a bounded kernel receive
+  buffer: trains are forwarded in chunks of at most 8 packets and
+  tail-dropped once the connection's mailbox holds more than 32
+  messages, keeping the head packet so the peer still gets a timely
+  ACK. An unbounded mailbox turned receiver overload into queueing
+  delay instead of loss, which inflated the peer's RTT samples and
+  destabilized its loss detector.
 
 ## [1.8.2] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,14 @@ All notable changes to this project will be documented in this file.
   gets a timely ACK. An unbounded mailbox turned receiver overload into queueing
   delay instead of loss, which inflated the peer's RTT samples and
   destabilized its loss detector.
+- With the crypto NIF loaded, a bulk chunk run is sealed in one call:
+  `protect_run` builds each short header, derives the nonce, AEAD-seals
+  the payload and applies header protection for the whole run in C,
+  instead of two crypto NIF calls plus header, nonce and mask glue per
+  packet. AES only (the header-protection context is ECB); ChaCha and
+  NIF-less builds seal per packet as before. Send-side crypto drops
+  from about 9% of connection CPU to about 2% on bulk flows. The output
+  is checked byte for byte against the per-packet path.
 
 ## [1.8.2] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,14 @@ All notable changes to this project will be documented in this file.
   NIF-less builds seal per packet as before. Send-side crypto drops
   from about 9% of connection CPU to about 2% on bulk flows. The output
   is checked byte for byte against the per-packet path.
+- With the crypto NIF loaded, a received 1-RTT packet is opened in one
+  call: `open_packet` does header unprotection, packet-number
+  reconstruction (RFC 9000 Appendix A) and AEAD open together, replacing
+  two crypto NIF calls plus mask, packet-number and AAD glue per
+  packet. It runs only while no key update is in flight and checks the
+  key-phase bit before decrypting, handing a phase change to the
+  two-stage path so RFC 9001 §6 bookkeeping stays where it was. ChaCha
+  and NIF-less builds are unchanged.
 
 ## [1.8.2] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,18 @@ All notable changes to this project will be documented in this file.
   their batched receive path engages even for sparse traffic. ACK
   processing re-arms the PTO once at the end of the receive pass
   instead of once per ACK frame.
+  processing re-arms the PTO once at the end of the receive pass
+  instead of once per ACK frame.
+- Server connections on the socket backend hand their finished send
+  batches to one shared sender process per listener, which performs
+  the `sendmsg` calls serially. Concurrent `sendmsg` on one socket
+  handle from many connection processes burns about 20x the CPU in
+  NIF-level contention (98.6 vs 4.7 us per send with 50 concurrent
+  senders); in a 50-connection fan-out model server CPU went from 24.3
+  to 11.0 s for the same window. Connections keep their own batch
+  buffers, GSO grouping and counters; a connection whose sender has
+  died sends directly from then on, and the listener starts a new one
+  for new connections.
 
 ### Fixed
 - Pacing no longer freezes on sub-millisecond links. RTT samples are
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,20 @@ All notable changes to this project will be documented in this file.
   one complete application message break when deliveries merge, so it
   is opt-in.
 
+### Fixed
+- Pacing no longer freezes on sub-millisecond links. RTT samples are
+  whole milliseconds, so such a link reports a smoothed RTT of 0 and
+  `update_pacing_rate` treated that as "no RTT yet" and skipped the
+  update: the rate stayed at its handshake-time value while cwnd kept
+  growing, clocking the connection at that stale rate forever. The RTT
+  is floored at 1 ms for the rate computation instead.
+- Burst continuations are sent to the connection directly rather than
+  through `erlang:send_after(0, ...)`. The timer wheel's ~1 ms service
+  tick turned every 64-packet burst continuation into a 64 packets per
+  millisecond clock, capping bulk streams at roughly 85 MB/s regardless
+  of the paced rate. Together with the previous fix, a verified
+  single-stream loopback bulk transfer went from 83 to 194 MiB/s.
+
 ### Changed
 - The interop runner declares the passive robustness cases (longrtt,
   blackhole, amplificationlimit, handshakeloss, transferloss,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ All notable changes to this project will be documented in this file.
   ACK. An unbounded mailbox turned receiver overload into queueing
   delay instead of loss, which inflated the peer's RTT samples and
   destabilized its loss detector.
+- Count-based ACK decimation defers its flush while a receive pass is
+  active, so one ACK covers the whole drained train instead of one per
+  `ack_packet_tolerance` packets. A 64-packet pass previously emitted
+  about 32 ACKs, each costing a packet build, an AEAD seal and a send
+  on the receiver and a decrypt plus ACK-frame pass on the sender; it
+  also released the sender's window in 2-3 packet quanta, which kept
+  GSO batches near size 1. The max_ack_delay timer still bounds ACK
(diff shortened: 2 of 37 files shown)


jbevemyr:
Rig numbers, as promised. Two physical hosts on 1 GbE (server: 32-core Ubuntu 24.04, OTP 28 in a container on host networking; client: OTP 28 native), RTT 0.2 ms, both NICs with generic GSO/GRO on and no hardware UDP segmentation. Same bench code as yours with the client half pointed at the remote host; `QUIC_DISABLE_NIFS=1` on both ends for the NIF-less rows. Three runs each, all with `status=ok` (byte count and CRC verified).

**Download** (server sends, 10 MB sink, the receive side you measured on loopback):

|                    | MB/s         | server retransmits |
|--------------------|--------------|--------------------|
| main               | 25.5 to 28.1 | 0 to 1             |
| series without NIF | 30.1 to 51.3 | 0                  |
| series with NIF    | 69.5 to 70.3 | 0                  |

**Upload** (client sends, 50 MB sink):

|                    | MB/s         | client retransmits |
|--------------------|--------------|--------------------|
| main               | 57.3 to 60.9 | 0                  |
| series without NIF | 56.8 to 73.5 | 25 to 82           |
| series with NIF    | 48.9 to 61.1 | 21 to 127          |

So on a real link the fused receive path holds up: 2.5x main on download, with the run-to-run spread of the pure-Erlang receiver gone. Upload is flat against main, and the NIF path does not help the sending side there (it was never expected to; the sealing gain is small next to the link).

Two things the rig found that loopback could not, both now in the series (force-pushed, same 25 commits):

1. A real ordering bug in the client mailbox drain, in "Open a GRO train of short-header packets in one NIF call". `collect_client_udp/5` matched only the single-datagram message shape, so a selective receive pulled those past earlier queued GRO trains and the connection saw the wire out of order. On loopback nearly everything arrives as trains so it never showed; on the link the mix of shapes made the receiver acknowledge a newer packet while a couple of hundred older ones sat unprocessed behind it, the sender declared them lost by packet threshold, retransmitted, and the storm fed itself: 1.6 to 14k retransmits per 10 MB, throughput down to 10 MB/s in the worst run. Both shapes are now taken in one receive, with a test that pins mailbox order. That is where the "series without NIF" download row above comes from; before the fix it was 10 to 52 MB/s with up to 14k retransmits.

2. The upload retransmits are real drops, and small: fq_codel on the client's NIC drops one or two GSO super-packets per 50 MB when the unpaced-by-timer sender bursts past its 5 ms target (2 qdisc drops for 53 retransmits in the run I sampled; 0 kernel receive drops and 0 listener drops on the server). What makes them expensive is PMTU black-hole detection: six large-packet losses without an ACK in between drop the path MTU to 1200 for the rest of the connection, so those runs send about 15% more packets (44k against 38.5k). That logic predates the series; I would like to make it count loss events rather than packets in a follow-up, unless you would rather have it in here.

jbevemyr:
One more pair on the same 1 GbE segment, with an idle 8-core i7-9700K (OTP 28 native) as the client against the 32-core server, three runs each, all verified:

|                    | upload, client sends 50 MB | download, client receives 10 MB |
|--------------------|----------------------------|---------------------------------|
| main               | 82.9 MB/s, 0 retransmits   | 40.4 to 49.7 MB/s, 0            |
| series without NIF | 110.2 to 110.3 MB/s, 0     | 54.5 to 68.1 MB/s, 0 to 283     |
| series with NIF    | 110.3 to 110.4 MB/s, 0     | 68.9 to 70.9 MB/s, 0            |

Upload from an idle sender is the link (110 MB/s is what 1 GbE delivers after headers) with zero retransmits, so the qdisc drops in the previous table were the loaded workstation's doing, not the sender's. The receive side again: the pure-Erlang receiver in the series is 1.3x main with some spread, the fused NIF path is 1.5x main and steady, and the one run with 283 retransmits was 83 kernel receive-buffer drops on the client (the pure-Erlang receiver falling behind a 1 GbE burst), which the NIF receiver did not have.

With the i7 as the server instead (workstation client), download main 21.9 to 24.9, series without NIF 28.1 to 38.9, series with NIF 46.5 to 53.9 MB/s, zero retransmits throughout.

jbevemyr:
And the cell that matters most for us: a Raspberry Pi 4 (four Cortex-A72, Debian 13 arm64, OTP 28 in a container on host networking) on the same gigabit segment, the kind of edge host this transport runs on. 20 MB up, 10 MB down, three runs each, verified. The Pi's kernel receive-buffer cap is at Debian's default (`net.core.rmem_max` 212992, so the bench's 7 MB `recbuf` is clipped) in the first table and raised to 8 MB in the second.

**Pi as the receiver** (fast peer sends into the Pi):

| rmem_max 212 KB    | Pi is the server (upload) | Pi is the client (download) |
|--------------------|---------------------------|-----------------------------|
| main               | 6.3 to 6.6 MB/s, 0 rtx    | 5.9 to 6.7 MB/s, 0 rtx      |
| series without NIF | 8.7 to 10.9, 114 to 213   | 11.4 to 12.1, 850 to 950    |
| series with NIF    | 15.8 to 16.7, 434 to 2287 | 13.4 to 14.0, 1023 to 1260  |

| rmem_max 8 MB      | Pi is the server (upload) | Pi is the client (download) |
|--------------------|---------------------------|-----------------------------|
| series without NIF | 14.1 to 14.9 MB/s, 0 rtx  | 14.7 to 16.0 MB/s, 0 rtx    |
| series with NIF    | 33.8 to 35.0, 0           | 19.6 to 20.1, 0             |

**Pi as the sender** (into the fast peer):

|                    | Pi is the server (download) | Pi is the client (upload) |
|--------------------|-----------------------------|---------------------------|
| main               | 11.0 to 11.4 MB/s, 0 rtx    | 13.3 to 14.4 MB/s, 0 rtx  |
| series without NIF | 16.3 to 16.7, 0             | 21.3 to 21.9, 0           |
| series with NIF    | 22.0 to 23.7, 0             | 31.9 to 32.4, 0           |

So on the weak host the fused path is worth 3x on receive (5x on the server-side receive path once the kernel buffer is not the limit) and 2 to 2.3x on send, with no loss anywhere. The retransmits in the first table are all kernel receive-buffer drops on the Pi (counted from `/proc/net/snmp` around each run, listener drops 0): the series' sender fills a 212 KB socket buffer in a couple of milliseconds, main only stays drop-free because its receiver ACK-clocks it down to 6 MB/s. With the buffer at 8 MB every drop and retransmit disappears, so that is a deployment setting for small hosts rather than a transport change. The upload/download asymmetry on the Pi is the listener process reading the socket on the server side against the connection process on the client side; the NIF receive path sits behind both.

jbevemyr:
Wrapping up the rig work in one place. Everything below is the series at a5bc383 against main 1.8.2, three verified runs per cell, upload = client sends (50 MB on x86, 20 MB on the Pi and the VMs), download = server sends (10 MB).

**Throughput, MB/s: main / series without NIF / series with NIF**

| client → server                     | upload                        | download                      |
|-------------------------------------|-------------------------------|-------------------------------|
| workstation → 32-core, 1 GbE        | 57–61 / 57–74 / 49–61         | 26–28 / 30–51 / 69–70         |
| idle i7 → 32-core, 1 GbE            | 83 / 110 / 110 (line rate)    | 40–50 / 54–68 / 69–71         |
| workstation → i7, 1 GbE             | 57–59 / 42–61 / 35–46         | 22–25 / 28–39 / 46–54         |
| i7 → Raspberry Pi 4, rmem 8 MB      | 6.3–6.6 / 14–15 / 34–35       | 11 / 16–17 / 22–24            |
| Raspberry Pi 4 → 32-core, rmem 8 MB | 13–14 / 21–22 / 32            | 6–7 / 15–16 / 20              |
| KVM guest → KVM guest, 2 vCPU, rmem 8 MB | 31–36 / 84–93 / 296–352  | 24–25 / 29–41 / 53–63         |

Retransmits are zero in every cell except the workstation-as-sender ones (its fq_codel drops one or two GSO super-packets per 50 MB, see above) and the Pi and VM rows before the kernel buffer was raised: with Debian's 212 KB default the faster NIF sender overruns a slow receiver's socket buffer, the pure-Erlang sender is slow enough not to (VM download with NIF: 15 to 35 MB/s with 4 to 50 kernel drops amplified into 100 to 1300 retransmits; with 8 MB, 53 to 63 and zero). The VM link is a host bridge, so its upload numbers are not bounded by a NIC.

**What the NIF is worth, against the same series without it**

- Receive side on fast x86: 1.3 to 1.5x, and it removes the run-to-run spread of the pure-Erlang receiver (30 to 50 % between identical runs, a few percent with the NIF).
- Receive side on the Pi: 2 to 2.3x, and 3 to 5x against main.
- Send side from a fast idle host: nothing, both are at line rate.
- Send side from the Pi: 1.5x (21–22 → 32), 2.3x against main. The sealing gain only shows where the CPU cannot saturate the link, which is the edge host case.
- Loss behaviour: unchanged. With 0.5 % netem loss on the VM link the series retransmits alike with and without the NIF; open_run gap and reorder shapes are now unit-tested.

So the ~2 % ceiling from the profile holds for a fast sender on a fast link, and nowhere else: the receive path is where the per-packet Erlang work was, and that is what the fused path removes.

**Things the rig turned up** (all in this PR or noted above): the receive-queue bound below the flow-control window, the client drain reordering the wire, PMTU black-hole detection tripping on one burst loss (follow-up proposed), and small hosts needing `net.core.rmem_max` above Debian's 212 KB. The 1-vCPU VM cell was scheduling noise (identical no-NIF runs swung 14 to 64 MB/s); the 2-vCPU row above is the one to read.

Raw logs, scripts and the per-cell tables with retransmit counts are on our side if you want any of them.

jbevemyr:
Superseded by #289, which carries these commits plus Benoit's build and review fixes.

PR #289, opened 2026-09-05
Performance series: interop cases, receive pass, delivery coalescing, send side, fused crypto NIF

benoitc:
Supersedes #288, whose 25 commits are carried over unchanged with jbevemyr's authorship. Five commits are added on top: the NIF build moves to CMake and gains macOS and FreeBSD support, and three findings from the review of #288 are fixed.

## Build

`c_src/Makefile` was GNU make and used `ldd`, so the NIF built on Linux only. It is now a CMake build that resolves the ERTS headers and the same libcrypto OTP's own crypto NIF is linked against, and emits the per-platform link flags itself. A thin POSIX `c_src/Makefile` shim stays for erlang.mk, which runs it unconditionally when it exists.

Verified under qemu on FreeBSD 14.3 arm64 (base OpenSSL) and locally on macOS arm64 (MacPorts OpenSSL 3): the NIF builds with no warnings, loads, and the full eunit suite passes both with it loaded and with `QUIC_DISABLE_CRYPTO_NIF=1` (2514 tests each). For reference, the old Makefile under FreeBSD's bmake:

```
make: "c_src/Makefile" line 13: Unknown modifier "format("~ts/erts-~ts/include", [code"
```

The build stays best effort everywhere: no cmake, no toolchain or no OpenSSL headers prints a note, succeeds, and the library runs on the OTP crypto path.

## Fixes

**A client Initial flight larger than one datagram was truncated.** The receive sweep groups a flight's packets by their unknown DCID, and the group that starts the connection forwarded only its first packet; the rest were dropped and the handshake stalled until the client retransmitted. A chunked `x25519mlkem768` ClientHello makes this the normal case, and the sweep put it on the default `gen_udp` path. The regression test freezes the listener so both Initials land in one sweep and asserts both reach the connection; with the fix reverted it sees one packet instead of two.

**The AEAD context cache grew without bound and could raise.** Every key update derives fresh keys and the cache kept a live `EVP_CIPHER_CTX` for each, so a long-lived bulk connection accumulated them until it exited. Only the current and previous key phase are ever live (RFC 9001 section 6), so the oldest are evicted. Separately, `{ok, Ctx} = quic_crypto_nif:new_aead_ctx(...)` was the one NIF call with no fallback: an unbuildable key killed the connection with a badmatch instead of taking the OTP path.

**The loader could make the module unloadable.** `code:priv_dir(quic)` answers `{error, _}` when the app is not resolvable, and passing that to `filename:join/2` crashed `on_load`, after which every caller died with `undef` rather than falling back. That defeats the point of an optional NIF.

**CI never proved the NIF was loaded.** The NIF tests skip themselves when it is missing, so a silent build failure would have turned six test modules into no-ops behind a green run. A new job asserts the shared object was built and loaded, then runs the suite with the NIF and with `QUIC_DISABLE_CRYPTO_NIF=1`.

Also documents the optional NIF, its two runtime opt-outs and the platforms it is verified on, plus `delivery_coalescing`, which shipped without an entry in the options reference.

Two limits worth stating: ChaCha20-Poly1305 connections keep only the per-packet acceleration, since the fused paths assume AES-ECB header protection, and the performance tables in #288 are Linux figures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,6 +258,55 @@ jobs:
           QUIC_DISABLE_CRYPTO_NIF: "1"
         run: rebar3 eunit
 
+  crypto-nif-fuzz:
+    name: Crypto NIF Fuzz (ASan)
+    needs: unit-tests
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Erlang
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: '28'
+          rebar3-version: '3.24.0'
+
+      - name: Install build dependencies
+        run: sudo apt-get update && sudo apt-get install -y cmake libssl-dev
+
+      # Compile first: the rebar pre_hook builds the NIF, and the ASan
+      # build below has to be the one left in priv/.
+      - name: Compile
+        run: rebar3 as test compile
+
+      - name: Rebuild the NIF with AddressSanitizer
+        run: |
+          rm -rf c_src/build priv/quic_crypto_nif.so
+          cmake -S c_src -B c_src/build \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
+            -DCMAKE_MODULE_LINKER_FLAGS="-fsanitize=address"
+          cmake --build c_src/build
+          test -f priv/quic_crypto_nif.so || { echo "NIF was not built"; exit 1; }
+
+      # open_run parses frames in C from plaintext the peer controls, so
+      # this is the one place attacker input reaches a hand-written
+      # parser. ASan aborts the emulator on a finding, which fails the
+      # step. Leak detection is off: the BEAM is not leak-clean and its
+      # allocations are not what is under test.
+      - name: Fuzz the in-NIF frame parser
+        env:
+          ASAN_OPTIONS: detect_leaks=0:abort_on_error=1
+        run: |
+          erlc -o _build/test/lib/quic/test test/quic_nif_fuzz.erl
+          LD_PRELOAD=$(gcc -print-file-name=libasan.so) \
+            erl -noshell -pa _build/test/lib/quic/ebin -pa _build/test/lib/quic/test -eval '
+              application:ensure_all_started(quic),
+              case quic_crypto_nif:is_loaded() of
+                false -> io:format("NIF not loaded, nothing fuzzed~n"), halt(1);
+                true -> quic_nif_fuzz:run(200000), halt(0)
+              end.'
+
   dist:
     name: Distribution Tests
     needs: unit-tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,25 +58,85 @@ jobs:
       - name: Run property tests
         run: rebar3 proper
 
-  # unit-tests-freebsd:
-  #   name: Unit Tests on FreeBSD
-  #   runs-on: ubuntu-latest
-  #
-  #   steps:
-  #     - uses: actions/checkout@v4
-  #
-  #     - name: Test on FreeBSD
-  #       uses: vmactions/freebsd-vm@v1
-  #       with:
-  #         release: '14.2'
-  #         usesh: true
-  #         prepare: |
-  #           pkg install -y erlang rebar3 git
-  #         run: |
-  #           erl -eval 'io:format("OTP ~s~n", [erlang:system_info(otp_release)]), halt().' -noshell
-  #           rebar3 compile
-  #           rebar3 eunit
-  #           rebar3 proper || true
+  macos:
+    name: macOS
+    needs: unit-tests
+    runs-on: macos-15
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # brew ships the current OTP; the point of this job is the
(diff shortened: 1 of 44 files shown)


jbevemyr:
Thanks, that is a better shape than what I could have done from outside: the CMake build was overdue, and the Initial-flight truncation is a real one that the sweep made reachable on the default path, good catch.

One small thing your snapshot predates: the last push to #288 (a5bc383, tests only) added two `open_run` cases for the shapes the rig made me worry about, a hole in the packet-number sequence inside a train and an older packet after newer ones. Both pass on this branch too; the hunk is below if you want it, otherwise I will send it as a follow-up once this lands.

<details><summary>test/quic_open_run_tests.erl</summary>

```diff
diff --git a/test/quic_open_run_tests.erl b/test/quic_open_run_tests.erl
index 37d2a80..8153a3a 100644
--- a/test/quic_open_run_tests.erl
+++ b/test/quic_open_run_tests.erl
@@ -122,3 +122,26 @@ open_run_empty_test() ->
             )
         )
     end).
+
+%% A run with a hole in the packet-number sequence (a lost datagram in
+%% the middle of a GRO train) must still open every present packet with
+%% its own packet number, carrying the largest forward across the gap.
+open_run_with_pn_gap_test() ->
+    with_nif(fun() ->
+        [A, B] = build_run(10, ?FB_BASE, payloads(2)),
+        [C, D] = build_run(13, ?FB_BASE, payloads(2)),
+        {ok, Results} = quic_aead_ctx:open_run(
+            aes_128_gcm, ?KEY, ?IV, ?HP, 9, 0, byte_size(?DCID), [A, B, C, D]
+        ),
+        ?assertEqual([10, 11, 13, 14], [PN || {PN, _FB, _P} <- Results])
+    end).
+
+%% Reordered within a train: an older packet after newer ones.
+open_run_reordered_test() ->
+    with_nif(fun() ->
+        [A, B, C] = build_run(20, ?FB_BASE, payloads(3)),
+        {ok, Results} = quic_aead_ctx:open_run(
+            aes_128_gcm, ?KEY, ?IV, ?HP, 19, 0, byte_size(?DCID), [B, C, A]
+        ),
+        ?assertEqual([21, 22, 20], [PN || {PN, _FB, _P} <- Results])
+    end).
```

</details>

Closing #288 in favour of this.

benoitc:
Taken, thanks. Instead of applying the hunk I rebased my five commits onto a5bc383, so this PR has your series as you pushed it, with mine on top and no duplicate commit to reconcile later. The net diff against main moved by exactly those 23 lines.

Both cases pass with the NIF loaded. `quic_open_run_tests` is 8 tests now, and the full suite is 2516 with the NIF and 2516 with `QUIC_DISABLE_CRYPTO_NIF=1`.

The PN-gap one is the good case. `open_run` carries the largest-received across the hole while it reconstructs each packet number from the truncated bits, so a datagram lost mid-train is where such a bug hides.

Two things I left out, yours to take or not. The fuzz harness is still not in CI, and it fuzzes a C parser reading attacker-controlled plaintext after decryption, so today it runs only when someone thinks about it. Doing it properly needs an ASan build. And ChaCha still declines the three fused paths, which matters more than the tables show: with `OPENSSL_armcap=0` here, AES-128 goes from 0.446 to 5.351 us per seal, ChaCha only from 1.024 to 1.912. Without hardware AES, ChaCha is 2.4x the faster cipher and gets none of the acceleration. Header protection is the blocker, its sample is the IV, so no context can be cached.

Also #290, based on main: it reuses the AEAD context on the pure-Erlang path (OTP 28+, one-shot below). Independent, but the same AEAD calls, so the second one to land needs a small merge.

jbevemyr:
With #290 in, this shows as conflicting; I looked at the merge so you do not have to guess at the shape. Two files. CHANGELOG is keep-both. `quic_aead.erl` has five hunks, all the same question: the series routes `encrypt`/`decrypt` through `quic_aead_ctx:aead_encrypt`/`aead_decrypt` (NIF when loaded, `crypto:crypto_one_time_aead` otherwise), #290 routes them through your `aead_seal`/`aead_open` (cached OTP 28 context, one-shot below).

The resolution that keeps both gains: take the series side at the five call sites, and point `quic_aead_ctx`'s two fallbacks at your helpers instead of the raw one-shot call, so a node without the NIF gets the context reuse and a node with it is unchanged:

```erlang
%% quic_aead_ctx.erl
fallback_encrypt(Cipher, Key, Nonce, Plaintext, AAD) ->
    quic_aead:aead_seal(Cipher, Key, Nonce, AAD, Plaintext).

fallback_decrypt(Cipher, Key, Nonce, Ciphertext, AAD, Tag) ->
    quic_aead:aead_open(Cipher, Key, Nonce, AAD, <<Ciphertext/binary, Tag/binary>>).
```

with `aead_seal/5` and `aead_open/5` exported from `quic_aead`. `aead_open` already returns `error` for a bad tag and for a too-short payload, which is what the ctx path's callers expect. The ChaCha per-packet path then also gets #290's reuse on OTP 28, which is the one cipher the fused paths do not cover.

Happy to push that as a commit on a branch of ours if you would rather cherry-pick than resolve by hand.

jbevemyr:
On the two you left open: we will take the ASan fuzz job as a follow-up once this lands (a CI job that builds the NIF with -fsanitize=address and runs quic_nif_fuzz), and treat ChaCha as a separate track. The armcap numbers make the case: the Pi 4 has no ARMv8 crypto, so that is exactly where we would want it. Neither should hold this PR.

PR #292, opened 2026-09-05
Run ChaCha20-Poly1305 through the fused crypto NIF paths

jbevemyr:
The ChaCha track from #289. ChaCha20-Poly1305 connections now go through the three fused NIF paths (`protect_run`, `open_packet`, `open_run`) like the AES suites.

The header-protection blocker was smaller than it looked. RFC 9001 5.4.4 defines the ChaCha mask as the first five keystream bytes of ChaCha20 keyed with the HP key, with the sample as counter and nonce. OpenSSL's `EVP_chacha20` takes exactly that as its 16-byte IV (little-endian counter, then nonce), so the NIF keeps a keyed ChaCha20 context per HP key and per packet does an IV reload and a five-byte update, the same shape as the AES-ECB context with an IV reload in place of a block. `new_hp_ctx` accepts `chacha20`, `hp_mask()` branches on the context kind, and the three fused functions call it instead of the ECB update inline. `hp_block/2` stays ECB-only and refuses a ChaCha context.

On the Erlang side `hp_cipher/1` maps the AEAD to its HP cipher and the three `chacha20_poly1305 -> fallback` clauses are gone. The per-packet Erlang path is unchanged, so a node without the NIF behaves as before.

Tests: the `protect_run` differential set now runs for ChaCha across the same packet-number boundaries as AES, `open_packet` roundtrips ChaCha runs, `open_run` has a ChaCha batch-versus-sequential case, and the RFC 9001 Appendix A.5 ChaCha packet is opened by the fused path and must yield PN 654360564 and the PING, which pins the C mask to the published vector rather than to our own Erlang. A new e2e case echoes 1 MB over a ChaCha connection; it passes with the NIF loaded and with `QUIC_DISABLE_CRYPTO_NIF=1`. Full suite, dialyzer, lint and the parser fuzz are green with and without the NIF.

No numbers in this one: the loopback boxes here have AES-NI, where AES stays ahead. The point is the host without it, per your `OPENSSL_armcap=0` measurement, and the Pi 4 on our rig is exactly that; I can post Pi numbers for ChaCha versus AES on it if useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,14 @@ All notable changes to this project will be documented in this file.
   socket, which is what caught this.
 
 ### Changed
+- ChaCha20-Poly1305 connections use the fused crypto NIF paths
+  (`protect_run`, `open_packet`, `open_run`) like the AES suites. The
+  header-protection mask is five ChaCha20 keystream bytes with the
+  16-byte sample as the cipher IV (RFC 9001 5.4.4), so the NIF keeps a
+  keyed ChaCha20 context per HP key and reloads only the IV per packet,
+  the same shape as the AES-ECB context. Where the CPU has no AES
+  instructions ChaCha is the faster cipher by a wide margin and was the
+  one suite left on the per-packet path.
 - The interop runner declares the passive robustness cases (longrtt,
   blackhole, amplificationlimit, handshakeloss, transferloss,
   handshakecorruption, transfercorruption, rebind-port, rebind-addr),
--- a/c_src/quic_crypto_nif.c
+++ b/c_src/quic_crypto_nif.c
@@ -25,7 +25,8 @@
 
 typedef struct {
     EVP_CIPHER_CTX *ctx;
-    int enc; /* 1 = seal, 0 = open */
+    int enc;     /* 1 = seal, 0 = open */
+    int hp_chacha; /* header-protection context: 1 = ChaCha20, 0 = AES-ECB */
 } qc_ctx;
 
 static ErlNifResourceType *qc_ctx_type;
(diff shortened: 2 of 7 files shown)


jbevemyr:
Numbers from the Raspberry Pi 4 (Cortex-A72, no ARMv8 crypto extensions, Debian 13 arm64, OTP 28 in a container) against the 32-core box over 1 GbE, this branch, `ciphers => [aes_128_gcm]` versus `ciphers => [chacha20_poly1305]` on both ends, fused NIF off and on. 20 MB up, 10 MB down, three runs, all verified. MB/s, min to max.

| Pi receives (fast peer sends) | AES-128 | ChaCha20 |
|---|---|---|
| per-packet path (NIF off), Pi is server | 14.4–14.6 | 15.4–16.1 |
| fused NIF, Pi is server | 32.2–35.5 | 50.8–57.1 |
| per-packet path (NIF off), Pi is client | 16.1–18.0 | 15.9–18.9 |
| fused NIF, Pi is client | 18.9–21.6 | 19.2–21.6 |

| Pi sends (fast peer receives) | AES-128 | ChaCha20 |
|---|---|---|
| per-packet path (NIF off), Pi is client | 24.6–25.2 | 27.8–30.1 |
| fused NIF, Pi is client | 31.5–33.4 | 55.2–58.1 |
| per-packet path (NIF off), Pi is server | 16.5–18.7 | 19.8–21.0 |
| fused NIF, Pi is server | 22.1–24.5 | 28.4–32.1 |

So on the host without hardware AES, ChaCha on the fused path is the fastest configuration by a clear margin: 1.6x AES on the same path both for the Pi receiving (server side) and for the Pi sending, and 3.5x the per-packet ChaCha path it was on before this PR. The client-side receive cell is flat across all four (19 to 22) because that path is bound elsewhere on the Pi. Zero retransmits and zero kernel drops in every cell (`rmem_max` at 8 MB on the Pi, as in the #288 tables).

PR #293, opened 2026-09-05
Build the ACK-isolation test state once per comparison

jbevemyr:
`quic_handshake_ack_isolation_tests` builds its state with the current clock in `time_sent`, and four tests build it twice, once for the before snapshot and once for the frame under test. Across a millisecond tick the two differ in that one field and the `assertEqual` fails on nothing the test is about; the macOS job on #292 hit it in `non_ack_frame_is_unaffected_test`. Each test now builds one state and uses it on both sides.


jbevemyr:
Hit twice more since: `empty_ack_is_a_no_op_at_every_level_test` on the macOS job for #294 (same double state() build, same tick boundary), so this fix covers all four tests, not just the one I saw first.

PR #294, opened 2026-09-06
Sweep the client socket before waking the connection

jbevemyr:
The last asymmetric cell from the #288 rig. On the Raspberry Pi 4 the two receive roles differed 2.5x with the same code, same peer and same direction of data: 51 to 57 MB/s into the Pi as server, 19 to 22 MB/s into the Pi as client, for every cipher and with or without the NIF.

A receive trace on the client connection explained it: 6330 single-datagram messages for a 10 MB download and not one train. GRO was enabled and useless, because nothing arrived back to back. The loop closes on itself: each datagram is its own message, so its own receive pass and its own ACK; the peer's next burst is sized to that ACK spacing (two packets); two packets spaced by the pacing never coalesce; and the receiver stays at one `recvmsg`, one `process_info` and one message per packet, which on the Pi is the whole budget. The listener never enters that loop because its receive sweep hands the connection up to 256 packets at a time, so one ACK covers a train and the peer's bursts grow until GRO takes over.

This gives the client receiver the same sweep at the socket: after the blocking read it drains what the socket already holds with non-blocking reads, up to 64 packets, and forwards the train as one message. A datagram from another source ends the sweep and is forwarded on its own after the train, so per-source order is kept. Nothing changes on the sending side or in the connection.

Same Pi, same peer (32-core box over 1 GbE), 10 MB download into the Pi as client, three runs, `rmem_max` 8 MB:

|                    | before, MB/s | with the sweep |
|--------------------|--------------|----------------|
| AES-128, NIF       | 18.9–21.6    | 32.3–33.6      |
| ChaCha20, NIF      | 19.2–21.6    | 34.0–40.7      |
| AES-128, no NIF    | 16.1–18.0    | 14.3–15.6      |
| ChaCha20, no NIF   | 15.9–18.9    | 16.1–16.9      |

Messages seen by the connection went from 6330 singles to 120 trains of 64 (plus a few smaller). The pure-Erlang rows do not move because the per-packet crypto is their bound; the NIF rows gain 1.6 to 1.9x and now sit close to the server-role cells. Uploads from the Pi are unchanged (the receiver only carries ACKs there).

Tests use real loopback sockets: a burst is returned in order as one train, a datagram from another source ends the sweep, the budget leaves the remainder on the socket for the next turn, and an empty socket returns the head alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,16 @@ All notable changes to this project will be documented in this file.
   socket, which is what caught this.
 
 ### Changed
+- The socket-backend client receiver sweeps the datagrams the socket
+  already holds before waking the connection, up to 64 per message, the
+  client-side twin of the listener's receive sweep. A paced sender's
+  small bursts used to reach the connection one datagram per message:
+  one receive pass and one ACK each, so the peer's next burst was sized
+  to that ACK spacing, GRO never saw two packets back to back and the
+  receiver stayed at one `recvmsg` per packet. On a Raspberry Pi 4
+  receiving a 10 MB download over 1 GbE the connection saw 6330
+  single-packet messages and 17 MB/s; with the sweep it sees 64-packet
+  trains and 46 MB/s, with the same code on the sending side.
 - The interop runner declares the passive robustness cases (longrtt,
   blackhole, amplificationlimit, handshakeloss, transferloss,
   handshakecorruption, transfercorruption, rebind-port, rebind-addr),
--- a/src/quic_socket.erl
+++ b/src/quic_socket.erl
@@ -69,6 +69,8 @@
 
 %% Idle tick of the shared sender loop; also keeps its receive bounded.
 -define(SHARED_SENDER_IDLE_MS, 30000).
+%% Packets per client receive sweep (matches the connection's drain cap).
+-define(CLIENT_RECV_SWEEP_MAX, 64).
 -include_lib("kernel/include/logger.hrl").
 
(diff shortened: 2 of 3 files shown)


jbevemyr:
Same A/B on fast hardware for completeness (32-core server → loaded workstation client over 1 GbE, 10 MB download, three runs): pure-Erlang receiver 30–51 MB/s before, 47–65 with the sweep; NIF receiver 69–70 before, 69–72 after. So where the receiver was already keeping up with the sender's rate the sweep changes nothing, and where the per-message overhead was the limit it helps; the Pi is the extreme of the second case.

jbevemyr:
Rebased on main after 1.10.0 (the changelog was the only conflict; the test now uses `get_socket/1` since `test_socket/1` moved). Re-measured on the same rig to see whether the per-space loss detection in #334 changed the picture on the client receive side: it does not, since `quic_socket` is untouched between 1.9.1 and 1.10.0 and the loop it closes is between ACK spacing and burst size, not loss recovery.

Raspberry Pi 4 as client, 10 MB download from a 32-core server over 1 GbE, AES-128 with the NIF, three runs, `rmem_max` 8 MB:

| | MB/s | ACKs sent |
|---|---|---|
| 1.9.1 | 18.5–20.2 | ~1 700 |
| 1.10.0 | 16.8–18.3 | ~1 600 |
| 1.10.0 + this PR | 36.6–41.6 | 21–72 |

Same 2x as before: without the sweep the connection still sees one datagram per message, with it the trains arrive and one ACK covers each.

PR #301, opened 2026-09-10
Receive path: client socket sweep, reassembly byte count, ACK-isolation test determinism

benoitc:
Aggregates #294, #295 and #293, which are all receive-path work and all merge cleanly together. Their commits are carried over by merge, not squash, so jbevemyr's authorship is intact. The three originals can be closed pointing here.

- **#294** sweeps the client socket before waking the connection, so a connection wakes once per sweep rather than once per datagram.
- **#295** keeps the stream reassembly byte count incrementally instead of walking the tree, changing the return shape of `keep_longest_chunk/3`, `trim_reassembly_buffer/2` and `extract_contiguous_data/2` to carry the delta.
- **#293** builds the ACK-isolation test state once per comparison, a test-determinism fix.

Verified the aggregate is exactly the sum of its parts: each PR commit is an ancestor of this branch, and the insertion count matches (171 + 121 + 16 = 308).

This is the first batch to go through the regression gate merged in #300, which asserts delivery, a retransmit ratio and a packets-per-byte ceiling rather than a throughput number.

Two of the three touch `src/quic_connection.erl`, so the crypto batch that follows will need one `git merge origin/main` for a CHANGELOG hunk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,16 @@ All notable changes to this project will be documented in this file.
   socket, which is what caught this.
 
 ### Changed
+- The socket-backend client receiver sweeps the datagrams the socket
+  already holds before waking the connection, up to 64 per message, the
+  client-side twin of the listener's receive sweep. A paced sender's
+  small bursts used to reach the connection one datagram per message:
+  one receive pass and one ACK each, so the peer's next burst was sized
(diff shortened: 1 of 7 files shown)


benoitc:
Superseded by #308.

PR #303, opened 2026-09-10
[bisect probe, do not merge] #294 alone against the regression gate

benoitc:
Throwaway probe. Batch #301 (#294 + #295 + #293) fails the regression gate on CI with 521 retransmits per 8,910 packets (5.8%) where main shows 6 per 7,710 (0.08%), while batch #302 passes the same gate on the same infrastructure. This isolates #294 to see whether the client socket sweep is the cause. Will be closed and the branch deleted once the gate result is in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,16 @@ All notable changes to this project will be documented in this file.
   socket, which is what caught this.
 
 ### Changed
+- The socket-backend client receiver sweeps the datagrams the socket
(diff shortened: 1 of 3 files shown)


benoitc:
Probe served its purpose: #294 alone fails the regression gate, so the client socket sweep is the cause rather than #295's reassembly accounting. Reproduced locally in Docker by constraining the socket buffer; details on #301.

PR #304, opened 2026-09-10
Receive path: reassembly byte count, ACK-isolation test determinism

benoitc:
Replaces #301, which bundled #294 as well. #294 fails the regression gate and is held back; see the analysis on #301. #295 and #293 are independent of it and carry the value.

- **#295** is the substantial one. The connection-level receive-buffer byte count was recomputed from the tree on every out-of-order stream frame, twice, walking every chunk. With one lost packet and a congestion window of data queued behind it, each further packet cost a walk of thousands of chunks: the receiver's mailbox grew into the thousands, ACKs went out seconds late, and the sender sat on its window. Measured by the author, a 50 MB download over 1 GbE with a single early loss ran at 13 MB/s where the same transfer without loss ran at 90 to 108. The stream now keeps a running count and every tree change reports its delta.
- **#293** builds the ACK-isolation test state once per comparison. Test determinism only, no runtime effect.

Both commits are carried by merge, so jbevemyr's authorship is intact.

Worth noting these matter more without #294, not less: #294's regression is that a constrained receiver drops packets, and #295 is precisely what stops a dropped packet from collapsing throughput.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,18 @@ All notable changes to this project will be documented in this file.
   for new connections.
 
 ### Fixed
+- A stream receiver with a hole at the head of its reassembly buffer no
+  longer slows down with every packet buffered behind it. The
+  connection-level receive-buffer byte count was recomputed from the
+  buffer on every out-of-order stream frame, twice, by walking every
+  chunk in the tree, so with one lost packet and a congestion window's
+  worth of data queued behind it each further packet cost a walk of
+  thousands of chunks, the receiver's mailbox grew into the thousands,
+  ACKs went out seconds late and the sender sat on its window: a 50 MB
+  download over 1 GbE with a single early loss ran at 13 MB/s where
+  10 MB, or a peer that never lost a packet, ran at 90 to 108. The
+  stream now keeps a running byte count and every tree change reports
+  its delta.
 - The NIF's AEAD context cache is bounded and no longer raises. Each
   key update derives fresh keys, and the per-process cache kept a live
   EVP context for every one of them, so a long-lived bulk connection
--- a/include/quic.hrl
+++ b/include/quic.hrl
@@ -598,6 +598,9 @@
     %% stop at the delivered point instead of visiting every buffered
     %% chunk, which matters once loss recovery has megabytes buffered.
     recv_buffer :: gb_trees:tree(non_neg_integer(), binary()),
+    %% Bytes held in recv_buffer, a running count so per-packet
+    %% accounting never walks the tree.
+    recv_buffered = 0 :: non_neg_integer(),
     %% Our recv side is terminal: FIN read (buffer empty) or peer RESET_STREAM.
     recv_done = false :: boolean(),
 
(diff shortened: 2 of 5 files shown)


jbevemyr:
Some data on the #294 verdict from #303, since the gate result turns out to depend on the receive buffer the runner happens to grant.

The upload harness asks for a 7 MB receive buffer, but the kernel clamps that to `net.core.rmem_max`. A GitHub runner has the Linux default of 212992, so the gate there runs with a 425984-byte socket buffer on both ends. On a host with `rmem_max` raised (64 MB here) the same request yields 14 MB, and there the gate passes main and every one of the eight open branches, #294 included, with 7703 to 7750 packets and zero retransmits per 10 MB.

Running the same upload with the buffer pinned to what the runner gets (`recbuf => 212992`, effective 425984), main and #294 interleaved, five samples each, 10 MB over loopback on a 16-core workstation:

| effective recbuf | main retransmits / packets | #294 retransmits / packets |
|---|---|---|
| 425984 | 231/9445, 613/9951, 661/9984, 286/8798, 697/10099 | 496/9556, 482/9850, 717/10118, 832/10172, 513/9792 |
| 851968 | 398/9568, 502/9781, 880/10035, 506/9755, 192/8918 | 177/8278, 255/8769, 682/9990, 297/9587, 313/8881 |
| 14 MB | 0/7728, 0/7724 | 0/7728, 0/7728 |

Library drop counters are zero throughout, so as in #303 it is the kernel dropping at the socket. Both branches are far over the 1 % bound at the runner's buffer size, and neither is consistently worse; the spread between runs of the same branch is larger than any difference between the branches. On the runner main lands at 0.08 % and #294 at 1.8 %, which fits the same picture with a slower host sitting near the edge: what the gate measures at that buffer size is how far the sender's bursts exceed a 426 KB receive buffer, and that is a sender-side pacing property main already has rather than something the client sweep introduces. The sweep can plausibly move the edge a little by delivering ACKs to the sender in trains, so the sender's window opens in bigger steps, but the burst it then emits is bounded by pacing, not by how the ACKs arrived.

Two things follow. The gate would be more reproducible if it set the buffer it measures at explicitly (`recbuf` is already an option of the harness) or the CI job raised `rmem_max` to what the harness asks for; as it stands the same code passes or fails depending on the host's sysctl. And the underlying sender behaviour at small receive buffers is worth a look of its own: with the buffer at the Linux default, a 10 MB upload over loopback retransmits 3 to 8 % on main, which is what a small host without tuned sysctls would see. Happy to dig into that one, it is the same pacing floor question I mentioned on #297.

None of this is an argument for merging #294 in this round; #304 without it is fine by me.

jbevemyr:
One more data point, from this PR's own CI: the Performance Regression job on the first run of #304 (https://github.com/benoitc/erlang_quic/actions/runs/34456593075/job/102805154748) failed the same assertion, 464 retransmits against 8562 packets on `upload_gen_udp`, library drop counters at zero, with #294 not in the branch. Same shape as the #303 probe, so the runner's socket buffer rather than the client sweep is what the gate is reacting to.

PR #341, opened 2026-09-22
Do not set SO_REUSEADDR on client sockets

jbevemyr:
With `SO_REUSEADDR` set, the kernel's ephemeral port allocation treats a port held by another `SO_REUSEADDR` socket as free. Two client sockets on one host can then autobind to the same port, and the kernel delivers every incoming datagram for that port to one of them. The other connection sends its ClientHello and its PTO probes, the server answers each one (its connection reaches `awaiting_client_finished`, 14 packets sent), and nothing reaches the client: `packets_received` stays 0 and the connection sits in `idle` until the idle timeout.

Found on a 1 GbE rig with 150 simultaneous `quic:connect` from one host to one server: about one connection per round stalled, on 1.9.1 and 1.10.0 alike, with the `socket` backend and with `gen_udp`. Dumping the 150 client sockets showed the stalled ones were exactly the ones sharing an ephemeral port with a sibling (e.g. 43227 and 51068 held by two sockets each in one round; all four stalled). Plain OTP sockets show the same kernel behaviour: 5 × 300 autobinds with `SO_REUSEADDR` gave 6 duplicate ports, 5 × 300 without gave none.

The fix removes the option from the client paths, `open_for_send` on both backends; listeners keep it for their fixed port so a restart can rebind, and `open/2` sets it only when the port is not 0.

Tests send one datagram first, since the port is assigned on the first send, then check that another `SO_REUSEADDR` socket cannot bind the client's port and that 200 clients get 200 distinct ports.

After the fix, 20 rounds × 150 simultaneous connects over 1 GbE: 0 duplicate ports, 0 stalls (before: 5 stalled rounds out of 12).


benoitc:
Superseded by #346, thanks for the diagnosis.

PR #355, opened 2026-09-23
Finish the HTTP/3 owner contract and accept a CA file

benoitc:
Six findings reported earlier and left unfixed. Two of them already broke a caller.

**The close event contradicted itself.** `closing/3` and `handle_quic_down/2` sent the owner a bare `closed`, `handle_quic_failed/2` sent `{closed, Reason}`, and `quic_h3:wait_connected/2` matches only the second, so a connection that died while connecting returned `{error, timeout}` after its full timeout. `docs/HTTP3.md` already documented the reason-carrying shape, and the QUIC layer underneath already sends one. Every route into `closing` now records why: `normal` for a local close or a drained GOAWAY, `owner_down` when the owner exits, `{h3_error, Code, Phrase}` for a protocol error, otherwise the QUIC reason.

**A stream this library reset was never reported.** Seven `quic:reset_stream/3` sites in `src/h3/`, none of which told the owner or the registered handler, so a body refused past `max_buffered_body`, a frame error, or `quic_h3:cancel/2` told the peer and left the local caller waiting. Four now report `{stream_reset, StreamId, ErrorCode}`, the event a peer reset already produces, and drop the stream's handler registration with it. The incomplete-request reset stays silent because it only runs inside the peer-reset path, which reports the stream once on its own.

**A finished request was never released, and a GOAWAY never drained.** Nothing removed a stream on a clean FIN. The half-close state was assigned rather than combined, so whichever FIN arrived last decided it and no stream reached `closed`. Since the GOAWAY drain waits for the stream map to empty, a connection that had served anything never closed gracefully. A stream reset is kept, so bytes still arriving on it are swallowed rather than read as a new stream.

**`cacertfile` was accepted nowhere** and silently ignored, so verification fell back to the OS trust store without saying so. It is read once where the connection or server starts. The three ad-hoc PEM readers move to one helper in `quic_cert`; the dist one matched a single certificate and crashed on a bundle.

`{error, send_queue_full}` is named in the specs and documented as the backpressure signal it is.

**On the upload path, I have to correct what I reported.** The 0.4 MiB/s against 93 MiB/s was not a like-for-like comparison. `quic_h3_bench:throughput/2` divided the bytes sent and the bytes received by the same round-trip duration, so its two figures were the same number by construction, and the 93 MiB/s came from the raw-QUIC harness, which sends in one write with 16 MB windows. Timed apart, 16 MiB through the echo server runs at 49 to 63 MB/s up against 44 to 57 MB/s down, over 256 sequential 64 KiB writes. There is no upload deficit, so I fixed the measurement and left the transport alone.

Downstream consumers of the close event need a one-line change; `quic:send_data/4` callers and anything matching a peer reset need none. Details in the branch notes.


benoitc:
Common Test caught a real consequence of dropping finished streams, and it is worth calling out.

`quic_h3_owner_SUITE:custom_owner_receives_datagram` sent an H3 datagram on a GET whose request and response had both finished. That only worked while finished streams were never released: `h3_send_datagram/3` uses the stream map purely as an existence check, so the leaked entry was what let it through.

Every real datagram session opens its request stream with `end_stream => false` and never finishes the request half, so it is never pruned. `erlang_masque` does this in all four of its session modules, and extended CONNECT does the same. I changed the case to hold the stream open the way those do, rather than weaken the pruning.

The behaviour change is in the CHANGELOG: `quic_h3:send_datagram/3` on a stream closed in both directions now returns `{error, unknown_stream}`.

Full `rebar3 ct` in the Linux container: 335 passed, 9 skipped, no failures.

